PR #8, opened 2021-02-09
feat: add docker-compose cloud configuration

gtirloni:
@greatislander This extracts the Dockerfile (and auxiliary files) from Sail and adds a separate docker-compose file useful only for cloud deployments. I feel like this avoids introducing incompatibilities with Sail so people can still use it and the instructions don't change.

I don't like working with Docker volumes because they are hard to backup so I introduced two new variables DB_DIR and REDIS_DIR to specify the path where the data directories should be mounted from.

I'm also not exposing MySQL or Redis (through FORWARD_ variables) because that's not necessary in the cloud deployment. If the time comes that we need to connect inside them to run ad-hoc commands, we just `docker exec` into them.

Please let me know if you have any suggestions. This largely follows the scheme used for the SJRK websites.

### DIFF
--- a/.github/workflows/dev-cleanup.yml
+++ b/.github/workflows/dev-cleanup.yml
@@ -30,10 +30,10 @@ jobs:
           proxy_port:     ${{ secrets.SSH_PROXY_PORT }}
           envs:     PROJECT_ID
           script: |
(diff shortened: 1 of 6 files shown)


greatislander (approved):
This looks great! I'll merge it.

PR #13, opened 2021-02-10
feat: configure backup system

greatislander:
WIP.


greatislander:
Merging as is— will revisit configuration later.

PR #208, opened 2021-11-09
feat: replace tabbed navigation with separate pages (resolves #204)

### LINKED ISSUE #204: Replace tabs with separate pages

greatislander:
After some testing we decided that tabbed navigation (ARIA tab pattern) doesn't scale well for entity, project and consultant pages. This PR replaces tabs with separate pages and secondary navigation for these pages.

- [x] Replace entity page tabs with separate pages
- [x] Replace project page tabs with separate pages
- [x] Replace consultant page tabs with separate pages

Resolves #204.


app/Http/Controllers/EntityController.php
```diff
@@ -60,9 +63,13 @@ public function store(CreateEntityRequest $request)
...
      * @return \Illuminate\View\View
      */
-    public function show(Entity $entity)
+    public function show(Entity $entity): View
     {
-        $entity->load('pastProjects', 'currentProjects');
+        if (Route::currentRouteName() === \locale() . '.entities.show-projects') {
+            $entity->load('currentProjects');
+        } elseif (Route::currentRouteName() === \locale() . '.entities.show-projects') {
+            $entity->load('pastProjects', 'currentProjects');
```
jobara, line 71: The two conditionals appear to be the same. 

Also I wonder if you could write this  as `request()->isRoute('entities.show-projects')`. You've done something similar in the blade templates.

See: 

- https://laravel.com/docs/8.x/requests#inspecting-the-request-path
- https://laravel.com/api/8.x/Illuminate/Http/Request.html#method_routeIs

greatislander: Good catch on the conditionals. Using `$request->isRoute()` means the `$request` variable would need to be set to a `Request` class instance representing the current request, which it isn't in that controller. I decided to use the Route facade instead. If you think there's a compelling reason to do so I could change it but it seemed like accessing the current request just to get the route name was more complex that checking it using the Route facade.

resources/views/entities/show.blade.php
```diff
@@ -1,58 +1,62 @@
...
-        </ul>
-
-        <div class="flow" id="about" x-bind="tabpanel">
-            <h2>{{ __('entity.section_about') }}</h2>
+    <div class="has-nav-secondary">
+        <nav class="secondary" aria-labelledby="entity">
+            <ul role="list">
+                <x-nav-link :href="localized_route('entities.show', $entity)" :active="request()->routeIs(locale() . '.entities.show')">{{ __('About') }}</x-nav-link>
+                <x-nav-link :href="localized_route('entities.show-accessibility-and-inclusion', $entity)" :active="request()->routeIs(locale() . '.entities.show-accessibility-and-inclusion')">{{ __('Accessibility and inclusion') }}</x-nav-link>
+                <x-nav-link :href="localized_route('entities.show-projects', $entity)" :active="request()->routeIs(locale() . '.entities.show-projects')">{{ __('Projects') }}</x-nav-link>
```
jobara, line 16: Where the route is defined it looks like it's just named `entities.show-projects`.  Is there somewhere that's updating the names with the locale and/or is it possible to set the `:active` state with just the name and not including the locale in the `routeIs` call?

greatislander: The actual route name with locale is retrieved using `localized_route`— basically it prepends a locale code. So the effective route name is `en.entities.show`, not `entities.show`. The locale does need to be included in the `routeIs` call as a result.

resources/views/consultants/show.blade.php
```diff
@@ -22,29 +22,36 @@
...
-            <li x-bind="tabWrapper"><a href="#about" x-bind="tab">{{ __('consultant.section_about') }}</a></li>
-            <li x-bind="tabWrapper"><a href="#interests-and-goals" x-bind="tab">{{ __('consultant.section_interests_and_goals') }}</a></li>
-            @can('update', $consultant)<li x-bind="tabWrapper"><a href="#lived-experience" x-bind="tab">{{ __('consultant.section_lived_experience') }}</a></li>@endcan {{--  TODO: handle this more granularly --}}
-            <li x-bind="tabWrapper"><a href="#professional-experience" x-bind="tab">{{ __('consultant.section_professional_experience') }}</a></li>
-            @can('update', $consultant)<li x-bind="tabWrapper"><a href="#access-needs" x-bind="tab">{{ __('consultant.section_access_needs') }}</a></li>@endcan {{--  TODO: handle this more granularly --}}
-        </ul>
-        <div class="flow" id="about" x-bind="tabpanel">
-            <h2>{{ __('consultant.section_about_person', ['name' => $consultant->firstName()]) }}</h2>
+    <div class="has-nav-secondary">
+        <nav class="secondary" aria-labelledby="project">
```
jobara, line 33: The `projects` id doesn't exist on an element in the consultants page.

greatislander: Resolved in effdf6b.

jobara:
Regarding the secondary navigation label. I'm not sure it's clear enough to only label it after the heading. Although I'm struggling to think of something that captures all the information that can be navigated to. Might be worth checking in with @lliskovoi and/or @cherylhjli to see if they have any better suggestions. 

Also the main navigation seems to be missing a label.

app/Http/Controllers/EntityController.php
jobara: When would these conditionals be hit?

greatislander:
> Regarding the secondary navigation label. I'm not sure it's clear enough to only label it after the heading. Although I'm struggling to think of something that captures all the information that can be navigated to. Might be worth checking in with @lliskovoi and/or @cherylhjli to see if they have any better suggestions.
> 
> Also the main navigation seems to be missing a label.

I've added "main menu" to the primary navigation. In terms of labelling the secondary menu my thinking was that since it will be announced as a navigation region, something like "Justin Obara navigation" or "Inclusive Design Research Centre navigation" would work well to clarify that you are navigating sections within those areas.

jobara (approved):

PR #212, opened 2021-11-17
feat: add translation management UI (resolves #209)

### LINKED ISSUE #209: Translation management interface for translatable fields

greatislander:
Adds translatable fields to the project creation and editing pages.

### Testing

1. Load the updated dev snapshot.
2. Log in as the entity user.
3. Navigate to http://localhost/en/entities/my-business/projects/create.
4. Create a new project, adding French translations where appropriate.
5. Edit the project.
6. Visit the project using both the English and French website interfaces.


app/Models/Project.php
```diff
@@ -78,6 +78,7 @@ class Project extends Model
      * @var array
      */
     protected $casts = [
+        'name' => 'array',
```
jobara, line 81: Is this an Array because of the JSON object stored in the DB for translations?

resources/lang/en/routes.php
```diff
@@ -6,4 +6,8 @@
     'login' => 'login',
     'register' => 'register',
     'verification.verify' => '/verify-email/{id}/{hash}',
+    'projects' => 'projets',
```
jobara, line 9: Because this is the English routes.php file, should this actually be:
```suggestion
    'projects' => 'projects',
```

resources/views/components/expander.blade.php
```diff
@@ -1,7 +1,7 @@
 <div class="expander" x-data="{expanded: false}">
     <x-header class="expander__summary" :level="$level">
-        <button x-bind:aria-expanded="expanded.toString()" x-on:click="expanded = !expanded">
-            {{ $summary }} <x-heroicon-s-plus x-show="! expanded" aria-hidden="true" class="icon" /><x-heroicon-s-minus x-show="expanded" aria-hidden="true" class="icon" />
+        <button type="button" x-bind:aria-expanded="expanded.toString()" x-on:click="expanded = !expanded">
+            {{ $summary }} <span class="badge">{{ __('Content added') }}</span> <x-heroicon-s-plus x-show="! expanded" aria-hidden="true" class="icon" /><x-heroicon-s-minus x-show="expanded" aria-hidden="true" class="icon" />
```
jobara, line 4: I believe initially you were going to have the "badge" be an aria-describedBy for the button instead of being part of the button. Was that idea changed, or did you mean to implement this like that?

greatislander: Oops, this is a case of me having edited the wrong file. Fixing now.

resources/lang/en/routes.php
greatislander: You're right, thanks!

app/Models/Project.php
greatislander: Yep, see: https://laravel.com/docs/8.x/eloquent-mutators#array-and-json-casting

resources/views/components/translatable-input.blade.php
```diff
@@ -0,0 +1,21 @@
...
+</div>
+@foreach($locales as $locale)
+    @if($locale !== locale())
+    <div class="expander field @error($name . '.' . $locale) field--error @enderror" x-data="{expanded: false, value: '{{ old($name . '.' . $locale, $model ? $model->getTranslation($name, $locale) : '') }}', badgeText: '{{ __('Content added') }}'}">
+        <p class="expander__summary" id="{{ Str::slug(__(':label (:locale)', ['label' => $label, 'locale' => get_locale_name($locale)])) }}">
+            <button type="button" x-bind:aria-expanded="expanded.toString()" x-on:click="expanded = !expanded" aria-describedby="{{ Str::slug(__(':label (:locale)', ['label' => $label, 'locale' => get_locale_name($locale)])) }}-status">
+                {{ __(':label (:locale)', ['label' => $label, 'locale' => get_locale_name($locale)]) }} <x-heroicon-s-plus x-show="! expanded" aria-hidden="true" class="icon" /><x-heroicon-s-minus x-show="expanded" aria-hidden="true" class="icon" />
+            </button>
+        </p>
+        <span class="badge" id="{{ Str::slug(__(':label (:locale)', ['label' => $label, 'locale' => get_locale_name($locale)])) }}-status" x-show="value && ! expanded" x-text="value ? badgeText : ''"></span>
```
jobara, line 14: Clicking on this element does not expand the input and sits over the button.

resources/views/components/translatable-input.blade.php
```diff
@@ -0,0 +1,21 @@
+<div class="field @error($name . '.' . locale()) field--error @enderror">
+    <x-hearth-label :for="$name . '_' . locale()" :value="__(':label (:locale)', ['label' => $label, 'locale' => get_locale_name(locale())])" />
+    <x-hearth-input type="text" :id="$name . '_' . locale()" :name="$name . '[' . locale() . ']'" :value="old($name . '.' . locale(), $model ? $model->getTranslation($name, locale()) : '')" />
```
jobara, line 3: I wonder if long string names like this would be easier to read using php string templates. Is that possible in the blade template?

jobara:
@greatislander when running the tests locally I get 6 skipped tests ( appear to be related to 2FA not being enabled ) and 1 failure.

```log
   WARN  Tests\Feature\TwoFactorAuthenticationTest
  - users must confirm password before enabling two factor authentication → Two-factor authentication support is not enabled.
  - users who have confirmed password can enable two factor authentication → Two-factor authentication support is not enabled.
  - users can authenticate with two factor code → Two-factor authentication support is not enabled.
  - users can not authenticate with invalid two factor code → Two-factor authentication support is not enabled.
  - users can authenticate with recovery code → Two-factor authentication support is not enabled.
  - users can not authenticate with invalid recovery code → Two-factor authentication support is not enabled.

  ---

  • Tests\Feature\ProjectTest > users with entity admin role can create projects
  Response status code [500] is not a redirect status code.
  Failed asserting that false is true.

  at tests/Feature/ProjectTest.php:42
     38▕         $url = localized_route('projects.show', ['project' => 'test-project']);
     39▕ 
     40▕         $response->assertSessionHasNoErrors();
     41▕ 
  ➜  42▕         $response->assertRedirect($url);
     43▕     }
     44▕ 
     45▕     public function test_users_without_entity_admin_role_cannot_create_projects()
     46▕     {


  Tests:  1 failed, 6 skipped, 121 passed
  Time:   378.72s
```

jobara:
@greatislander when I'm trying to run this locally and create a new project with the entity user, I get the following error:

```log
BadMethodCallException
Method Illuminate\Validation\Validator::validateUniqueTranslation does not exist. 
```

This happens even when trying to create a project with just an English project name.

greatislander:
@jobara For the tests, skipped tests are fine— we aren't using two-factor auth on this project so that's expected. Tests pass for me and in CI so I expect this and the `validateUniqueTranslation` error are both related to the fact that you need to run `sail composer install` to add a new dependency (https://github.com/accessibility-exchange/platform/blob/feat/translation-ui/composer.json#L14). I forgot to add this step to the testing instructions. Can you do that and run the test suite again (and also create a new project)?

jobara:
Updating composer removed the `BadMethodCallException` and the test failure. 

I also noticed some issues with setting the start/end date. 

1. It appears you can set an end date that occurs before the start date. I don't see where the end date is uses, so maybe it's just removed/ignored in these cases.
2. If you set a start date in January, it will return an error that the start date isn't provided, unless you first change to a different month and then back to January.

resources/views/components/translatable-input.blade.php
greatislander: Resolved in bffffeb.

resources/views/components/translatable-input.blade.php
greatislander: It appears that in component attributes you can't use string interpolation. At least, I've had no luck doing so…

greatislander:
> I also noticed some issues with setting the start/end date.
> 
> 1. It appears you can set an end date that occurs before the start date. I don't see where the end date is uses, so maybe it's just removed/ignored in these cases.
> 2. If you set a start date in January, it will return an error that the start date isn't provided, unless you first change to a different month and then back to January.

Thanks for catching this. I'd like to address it separately as there's other issues focused on project editing which involve updating/adding other validation methods to a number of fields.

resources/css/components/_expander.scss
```diff
@@ -1,6 +1,10 @@
+.expander {
+    position: relative;
+}
+
 .expander button {
     align-items: center;
-    background-color: var(--ll-background-color);
+    background-color: transparent;
```
jobara, line 7: If/when you integrate UIO, you'll need to remember to account for this when adding the contrast themes.

I also wonder if you could have also achieved the same interaction by setting the `pointer-events: none` on the badge. In a quick trial both seemed to behave similarly. However, there may be other issues that I didn't consider.

resources/views/components/translatable-input.blade.php
jobara: Thanks for checking.

jobara (approved):

PR #242, opened 2021-11-29
feat: rename consultants to community members (fix #241)

### LINKED ISSUE #241: Consultant: rename to community member

greatislander:
In our last round of co-design we reached the conclusion that rather than all members of the platform being identified as `consultants`, they  should be `community members` who can either participate or consult on accessibility projects. This PR makes some underlying database changes and front end (view) changes to make that terminological change throughout the application.


app/Models/Project.php
```diff
@@ -371,11 +371,11 @@ public function getEntitySubsteps(): array
     }
 
     /**
-     * Get the project substeps for consultants.
+     * Get the project substeps for community-members.
```
jobara, line 374: should this be participants. It seems that you refer to community members as participants in relation to the Project.

app/Models/Project.php
```diff
@@ -669,67 +669,67 @@ public function communities(): BelongsToMany
     }
 
     /**
-     * The consultants that are interested in the project.
+     * The participants that are interested in the project.
      *
      * @return \Illuminate\Database\Eloquent\Relations\BelongsToMany
      */
-    public function interestedConsultants(): BelongsToMany
+    public function interestedCommunityMembers(): BelongsToMany
```
jobara, line 676: I'm not sure if the comment or the function name should be updated, but it's probably better if the match. (i.e. that is participant vs community member). Is the concept that "participant" is used to refer to "community members" in the context of a project, or that "community members" become "participants" once they are added to a project? Or maybe something else.

greatislander: > "community members" become "participants" once they are added to a project

Exactly this.

resources/views/projects/entity-step/4.blade.php
```diff
@@ -11,28 +11,28 @@
...
     </div>
     @endif
-    @if(count($project->confirmedConsultants) > 0)
-    <h3>{{ __('Consultants') }}</h3>
-    @foreach($project->confirmedConsultants as $consultant)
-    <x-consultant-card :consultant="$consultant" level="4">
-        <x-heading :level="5">{!! __('<span class="visually-hidden">:name’s </span>Contact information', ['name' => $consultant->name]) !!}</x-heading>
+    @if(count($project->confirmedParticipants) > 0)
+    <h3>{{ __('Participants') }}</h3>
+    @foreach($project->confirmedParticipants as $communityMember)
```
jobara, line 16: Are you using `$communityMember` here instead of `$participant` to identity what "type" is being acted upon?

greatislander: Yep.

jobara (approved):

PR #687, opened 2022-08-04
Feat/create scheduler

marvinroman:
Added new command `deploy:initial` which is where we will put everything that we need to run at boot time. Then we only add this command into entrypoint script and make all changes within this command.

Made update to entrypoint script.

Added new command `db:refresh` which chains `migrate:fresh` & `db:seed` commands.

Added schedule to run command `db:refresh` to run nightly on environments tagged with **staging** or **local**.

### DIFF
--- a/app/Console/Commands/DatabaseRefresh.php
+++ b/app/Console/Commands/DatabaseRefresh.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class DatabaseRefresh extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'db:refresh';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Refreshes and seeds the database for development work.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $this->call('migrate:fresh', ['--force']); 
+        $this->call('db:seed', ['DevSeeder', '--force']);
+    }
+}
--- a/app/Console/Commands/DatabaseRefresh.php
+++ b/app/Console/Commands/DatabaseRefresh.php
@@ -27,7 +27,7 @@ class DatabaseRefresh extends Command
      */
     public function handle()
     {
-        $this->call('migrate:fresh', ['--force']); 
+        $this->call('migrate:fresh', ['--force']);
         $this->call('db:seed', ['DevSeeder', '--force']);
     }
 }
--- a/app/Console/Commands/DeployInitial.php
+++ b/app/Console/Commands/DeployInitial.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+
+class DeployInitial extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'deploy:initial';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'All commands that should be run when a container boots.';
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        // split migrate commands between production and development versions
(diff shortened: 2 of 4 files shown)


app/Console/Kernel.php
```diff
@@ -24,7 +24,10 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        // $schedule->command('inspire')->hourly();
+        $schedule->command('db:refresh') // use custom command to make sure that the commands are chained
+            ->daily() // Run daily at midnight
+            ->environments(['staging', 'local']) // only run for APP_ENV tagged staging or local
```
greatislander, line 29: The current dev deploy has `dev` as its environment.

```suggestion
            ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
```

.github/workflows/refresh-dev.yml
```diff
@@ -25,8 +25,7 @@ jobs:
           proxy_port: ${{ secrets.SSH_PROXY_PORT }}
           script: |
             cd /srv/tae-dev && \
-            /usr/local/bin/docker-compose -f docker-compose.cloud.yml exec -T app ./artisan migrate:fresh --force && \
-            /usr/local/bin/docker-compose -f docker-compose.cloud.yml exec -T app ./artisan db:seed DevSeeder --force
+            /usr/local/bin/docker-compose -f docker-compose.cloud.yml exec -T app ./artisan db:refresh
```
greatislander, line 28: If we add the scheduler to the dev environment, could we run `artisan schedule:work` here instead?

greatislander (changes requested):
A couple of suggestions/questions. Also wondering where the schedule is run on the staging environment?

app/Console/Commands/DatabaseRefresh.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        // don't run command in production or testing environment
+        if (in_array(config('app.env'), ['testing', 'production']) !== true) {
+            $this->call('migrate:fresh', ['--force' => true, '--step' => true]);
```
greatislander, line 32: What is the purpose of the `--step` flag here?

app/Console/Commands/DeployInitial.php
```diff
@@ -0,0 +1,46 @@
...
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        // split migrate commands between production and development versions
+        if (config('app.env') === 'production') {
+            $this->call('migrate', ['--step' => true, '--force' => true]);
+        } else {
+            $this->call('db:refresh');
```
greatislander, line 34: This won't do anything in testing because of the conditional check here: https://github.com/accessibility-exchange/platform/blob/ca5f89b0803377ce491c771b771cb5f138ba9057/app/Console/Commands/DatabaseRefresh.php#L31

app/Console/Commands/DatabaseRefresh.php
marvinroman: The `--step` might be unnecessary in fresh migrations. I usually like to run migrations as stepped. Meaning each is run separately and can be rolled back one at a time rather than as a large block.

.github/workflows/refresh-dev.yml
marvinroman: `artisan schedule:work` doesn't exit. 

I would probably launch a cron every minute like `/usr/local/bin/docker-compose -f docker-compose.cloud.yml exec -T app ./artisan schedule:run`.

marvinroman:
> Also wondering where the schedule is run on the staging environment?

Currently we aren't running a cron to hit scheduler in staging. We would need to open an issue and have Jure implement one into the Kube.

app/Console/Commands/DeployInitial.php
marvinroman: Should we remove these tests altogether?

greatislander: I think so @marvinroman.

marvinroman:
Reverted workflows, the changes seemed to cause testing failures.
Removed `--step` from `migrate:fresh`. 
Removed test for commands.

greatislander (approved):

PR #688, opened 2022-08-04
chore: local deployment for development

marvinroman:
If we add in dev it will make it easier to do local development and testing.


greatislander (commented):
Thanks for working on this, @marvinroman! How does this Docker configuration handle local mail delivery? I currently use [Mailhog](https://github.com/mailhog/MailHog) with Valet to receive email from the app during local development. Is something like this supported or could it be?

marvinroman:
Mailhog was left alone in the configuration. It should work the same as usual.

The changes where mostly limited to the container running Laravel so that it would mirror what's running within the Kube setup.

`docker compose up -d` will launch all the containers including **meilisearch**, **mailhog**, **selenium** as well.

marvinroman:
@greatislander I will close this pull request for now and use for local development for while before doing a pull request again for it.

PR #769, opened 2022-08-11
ci: split deploy commands into deploy:global and deploy:local

greatislander:
This PR splits the `deploy` commands into two commands:

- `deploy:local` is meant to be run on every webhead and handles storage linking and font downloads
- `deploy:global` is meant to be run on a single webhead and handles database and file caching

### DIFF
--- a/.deploy/accessibility-app/entrypoint.sh
+++ b/.deploy/accessibility-app/entrypoint.sh
@@ -22,11 +22,12 @@ then
 
   touch $FILES_PATH/../deploy.lock
 
-  php artisan deploy:shared
-  php artisan deploy:unique
+  php artisan deploy:global
 
 fi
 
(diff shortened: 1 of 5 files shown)


.deploy/accessibility-app/entrypoint.sh
```diff
@@ -22,8 +22,9 @@ then
 
   touch $FILES_PATH/../deploy.lock
 
-  php artisan deploy:initial
-  
+  php artisan deploy:shared
+  php artisan deploy:unique
```
JureUrsic: @greatislander  php artisan deploy:unique needs to be below line: 30, whatever is within "if [ ! -f $FILES_PATH/../deploy.lock ]" if statement will only run on one webhead

JureUrsic (approved):
Looks good

PR #901, opened 2022-08-22
feat: manage engagements, part one

greatislander:
Adds management steps for engagements as follows:

- [x] Choose and edit page translations
- [x] Confirm and edit selection criteria
- [x] Move participant numbers to selection criteria

### Prerequisites

- [x] I've run `npm run build` and committed built assets to the PR branch.


cherylhjli:
- Selection criteria
    - Location
        - When specific provinces or territories is selected, if I uncheck all the provinces and territories, it can still save without an error
        - When specific cities or towns is selected, if I don’t select any city or town, it can still save without an error
        - If there is an error with any other field on the page, the location goes back to specific provinces and territories even if I entered specific cities and towns. If I select cities and towns radio button, it does load up any locations I’ve previously inputted. Desired behaviour is for the specific cities and towns radio button to remain selected, and to show any entered cities and towns below.
    - Disability or Deaf groups
        - If I select “yes I’m interested in a specific disability or Deaf group” but I don’t check off any disabilities in the following question, and save, I get an error for not inputting a specific disability group at the top. However, when I scroll down to the specific question, the radio button selection goes back to “no, I’m not interested in a cross disability group”. If I select “yes, I’m interested…” again, the field with an inline error shows up. Desired behaviour - have the “yes, I’m interested…” remain selected and for the question “what specific disability and deaf groups are you interested in engaging?” Remain visible, so that users don’t miss the inline error.
    - Intersectionality
        - Same issue as the one mentioned under disability or deaf groups - if I select yes and save without selecting a specific type of experience or identity, it produces an error at the top of the page. But, the radio goes back to “no”. If I select “yes” again then it shows the inline error.
        - This issue is the same if I select any type of experience, like age, but don’t enter a specific age group - if I save, the selection goes back to “no” and I have to click “yes” and pick “age” under the dropdown to see the inline error. Desired behaviour is that this just stays open so the user can see the inline error
        - If I pick “age” and specify specific age brackets, but I don’t fill in the minimum and ideal participants and I save it and it produces an error, I come back and it didn’t save my age selection. I haven’t tested this with the other identities/experiences but it might be the same behaviour
    - When I’m editing my selection criteria, I can enter 0 for both minimum and ideal number and it doesn’t produce an error
    - UI
        - I realize that we had a conversation earlier about the grey blocks only appearing for specific cities and towns, and not having it under the other disclosure buttons (edit location criteria, etc). I think that was a bad call on my part - when using this I had a hard time telling what was expanded and what wasn’t. Could we have the grey blocks back for anything under the disclosure buttons? For the city and town blocks, we can use grey 1 for the background again.

PR #986, opened 2022-08-31
fix: use default working languages when the property is not specified (resolves #865)

### LINKED ISSUE #865: foreach() argument must be of type array|object, null given
**`ErrorException`**

foreach() argument must be of type array|object, null given

:dart: Illuminate\Foundation\Bootstrap\HandleExceptions::handleError
:page_facing_up: /app/resources/views/individuals/partials/about.blade.php

```php
10      @foreach($individual->working_languages as $language)
11          <li>{{ get_language_exonym($language) }}</li>
12*     @endforeach
13  </ul>
14  
```
[Open in Flare](https://flareapp.io/errors/1494987/latest)

chosww:
See issue #865 for details.

### DIFF
--- a/resources/views/individuals/partials/about.blade.php
+++ b/resources/views/individuals/partials/about.blade.php
@@ -11,9 +11,12 @@
             <li>{{ get_language_exonym($language) }}</li>
         @endforeach
     @else
-        @foreach ($individual->languages as $language)
-            <li>{{ get_language_exonym($language) }}</li>
-        @endforeach
+        <li>{{ get_language_exonym($individual->user->locale) }}
+            @if ($individual->user->singed_language)
+                @foreach ($individual->user->singed_language as $language)
+        <li>{{ get_language_exonym($language) }}</li>
+    @endforeach
+    @endif
     @endif
 </ul>
 
--- a/resources/views/individuals/partials/about.blade.php
+++ b/resources/views/individuals/partials/about.blade.php
@@ -13,9 +13,7 @@
     @else
         <li>{{ get_language_exonym($individual->user->locale) }}</li>
         @if ($individual->user->singed_language)
-            @foreach ($individual->user->singed_language as $language)
-                <li>{{ get_language_exonym($language) }}</li>
-            @endforeach
+            <li>{{ get_language_exonym($language) }}</li>
         @endif
     @endif
 </ul>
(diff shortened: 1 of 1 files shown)


resources/views/individuals/partials/about.blade.php
```diff
@@ -6,9 +6,15 @@
...
 <ul>
-    @foreach ($individual->working_languages as $language)
-        <li>{{ get_language_exonym($language) }}</li>
-    @endforeach
+    @if ($individual->working_languages)
+        @foreach ($individual->working_languages as $language)
+            <li>{{ get_language_exonym($language) }}</li>
+        @endforeach
+    @else
+        @foreach ($individual->languages as $language)
```
greatislander: These are actually the languages into which the individual can translate their page. We actually want to retrieve the users's `locale` and (if set) `signed_language` attribute.

greatislander: See: https://github.com/accessibility-exchange/platform/blob/3d7a134f0c7b9f6681ee53ed1568743e80c63dbf/app/Http/Controllers/IndividualController.php#L115-L121

greatislander (changes requested):

resources/views/individuals/partials/about.blade.php
```diff
@@ -11,9 +11,12 @@
...
     @else
-        @foreach ($individual->languages as $language)
-            <li>{{ get_language_exonym($language) }}</li>
-        @endforeach
+        <li>{{ get_language_exonym($individual->user->locale) }}
+            @if ($individual->user->singed_language)
+                @foreach ($individual->user->singed_language as $language)
+        <li>{{ get_language_exonym($language) }}</li>
+    @endforeach
+    @endif
```
greatislander: There will only be one so there's no need for a loop— also, there's a typo (signed not singed). Thanks @chosww!

greatislander (changes requested):
One last adjustment!

greatislander (approved):

resources/views/individuals/partials/about.blade.php
chosww: My bad, I kept making a small mistakes and had to commit small changes multiple times. I have addressed all the issues from your comments.

PR #1049, opened 2022-09-07
ci: fix deployment

gtirloni:
When we run docker-compose exec, it runs as the container's default user (root), but the sail's Dockerfile contains logic to run the app as the `sail` user (either through gosub or supervisord). Normally, people user Dockerfile's USER instruction but the sail author decided to do it this way (probably to make using the Dockerfile for development easier).

In this PR, I change the docker-compose commands so they run with `--user sail` to avoid permission errors and moved them into a separate shell script for better maintainability.

It also defines the `WWWUSER` env var as expected by the `start-container` script so that the `sail` UID is changed to that value. That value is the same as the host's deploy user's UID. So the deployment steps running outside the container (e.g. git pull) can change files and then the deployment steps inside the container (e.g. npm install) also can do their job.

Another approach would be to run the deployment steps at build time inside the Dockerfile. This is how it's usually done for production images but it'd mean we would diverge a lot from sail's Dockerfile and would need to keep it up to date ourselves. Today, it's much easier to just copy it from the `docker/8.1` folder and make small adjustments... so I'm keeping it that way.

### Prerequisites


bin/deploy-dev.sh
```diff
@@ -0,0 +1,28 @@
...
+git checkout -f
+git checkout dev
+git pull
+
+# Build image and (re)create containers
+$COMPOSE_CMD up -d --force-recreate --build
+
+# Deploy app
+$EXEC_CMD npm install
+$EXEC_CMD composer install --optimize-autoloader
```
greatislander, line 15: After assets are built, we should run this to put the site in maintenance mode:

```suggestion
$EXEC_CMD ./artisan down --render="maintenance"
$EXEC_CMD composer install --optimize-autoloader
```

bin/refresh-dev.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -ex
+
+CONTAINER_USER="sail"
+COMPOSE_CMD="/usr/local/bin/docker-compose -f docker-compose.cloud.yml"
+EXEC_CMD="$COMPOSE_CMD exec -T --user $CONTAINER_USER app"
+
+$EXEC_CMD ./artisan migrate:fresh --force
+$EXEC_CMD ./artisan db:seed DevSeeder --force
```
greatislander, line 11:
```suggestion
$EXEC_CMD ./artisan down --render="maintenance"
$EXEC_CMD ./artisan migrate:fresh --force
$EXEC_CMD ./artisan db:seed DevSeeder --force
$EXEC_CMD ./artisan up
```

bin/deploy-dev.sh
```diff
@@ -0,0 +1,28 @@
...
+$EXEC_CMD ./artisan migrate:fresh --force
+$EXEC_CMD ./artisan db:seed DevSeeder --force
+$EXEC_CMD ./artisan view:clear
+$EXEC_CMD ./artisan storage:link
+$EXEC_CMD ./artisan google-fonts:fetch
+$EXEC_CMD ./artisan icons:clear
+$EXEC_CMD ./artisan icons:cache
+$EXEC_CMD ./artisan route:cache
+$EXEC_CMD ./artisan config:cache
```
greatislander, line 24:
```suggestion
$EXEC_CMD ./artisan config:cache
$EXEC_CMD ./artisan up
```

greatislander (changes requested):
@gtirloni Thank you! I just added some suggestions to restore the commands which will put the site in maintenance mode during deploys and refreshes.

greatislander (approved):

PR #1081, opened 2022-09-22
feat(individual)!: check that page is publishable (Resolves #594)

### LINKED ISSUE #594: Individual: check that page is publishable

jobara:
Ensures that required fields in an Individual's model have been filled in before allowing allowing the isPublishable method to return true. Also uses this method to only show the "Publish" button for individual pages after the criteria has been met.

Resolves #594 

### Prerequisites

### DIFF
--- a/app/Policies/IndividualPolicy.php
+++ b/app/Policies/IndividualPolicy.php
@@ -63,13 +63,9 @@ class IndividualPolicy
 
     public function publish(User $user, Individual $individual): Response
     {
-        if ($user->id !== $individual->user_id) {
-            return Response::deny(__('You cannot publish this individual page.'));
-        }
-
-        return $individual->isPublishable()
+        return $user->id === $individual->user_id && $individual->isPublishable()
             ? Response::allow()
-            : Response::deny(__('This individual page is missing some required information, and cannot be published yet.'));
+            : Response::deny(__('You cannot publish this individual page.'));
     }
 
     public function unpublish(User $user, Individual $individual): Response
--- a/app/Policies/IndividualPolicy.php
+++ b/app/Policies/IndividualPolicy.php
@@ -63,9 +63,13 @@ class IndividualPolicy
 
     public function publish(User $user, Individual $individual): Response
     {
-        return $user->id === $individual->user_id && $individual->isPublishable()
+        if ($user->id !== $individual->user_id) {
+            return Response::deny(__('You cannot publish this individual page.'));
+        }
+
+        return $individual->isPublishable()
             ? Response::allow()
-            : Response::deny(__('You cannot publish this individual page.'));
+            : Response::deny(__('This individual page is missing some required information, and cannot be published yet.'));
     }
 
     public function unpublish(User $user, Individual $individual): Response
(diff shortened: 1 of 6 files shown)


jobara:
@greatislander I did add a test for what I believed should test the policy change, but the test coverage was still reporting a line being missed. Perhaps something else is blocking it as well? Any thoughts?

tests/Feature/IndividualTest.php
```diff
@@ -756,6 +774,52 @@
...
+test('individual pages cannot be published by other users', function () {
+    $user = User::factory()->create();
+    $individual = Individual::factory()->create([
+        'roles' => ['consultant'],
+        'published_at' => null,
+    ]);
+
+    $response = $this->actingAs($user)->from(localized_route('individuals.show', $individual))->put(localized_route('individuals.update-publication-status', $individual), [
+        'publish' => true,
+    ]);
```
jobara: @greatislander I tried removing the `from` portion, as per our conversation but it still wasn't hitting the publish policy. I believe the `put` will actually fail on the `update` policy from the 'individuals.update-publication-status' route. So I'm not sure it's actually possible to hit the line of code. I'm going to collapse the policy back down and for now revert back to the old message as it's more generic in case something does slip through.

app/Policies/IndividualPolicy.php
```diff
@@ -63,9 +63,13 @@ public function update(User $user, Individual $individual): Response
...
     {
-        return $user->id === $individual->user_id && $individual->isPublishable()
+        if ($user->id !== $individual->user_id) {
+            return Response::deny(__('You cannot publish this individual page.'));
+        }
+
+        return $individual->isPublishable()
             ? Response::allow()
-            : Response::deny(__('You cannot publish this individual page.'));
+            : Response::deny(__('This individual page is missing some required information, and cannot be published yet.'));
```
jobara: It doesn't appear that it is actually possible to hit the first condition. I've tried to write tests for it, but it gets caught on other policies first. I'm planning to revert this change to keep the more generic message just in case something does slip through.

resources/views/individuals/partials/progress.blade.php
```diff
@@ -23,7 +23,8 @@
         @if ($individual->checkStatus('draft'))
             <p class="stack">
                 <button class="secondary" name="preview" value="1">{{ __('Preview page') }}</button>
-                <button class="secondary" name="publish" value="1">{{ __('Publish page') }}</button>
+                <button class="secondary" name="publish" value="1"
+                    @cannot('publish', $individual) disabled @endcannot>{{ __('Publish page') }}</button>
```
greatislander, line 27: I was thinking following our earlier conversation, we'll need to address this but probably should do so in a separate PR to keep this one focused.

tests/Datasets/IndividualIsPublishable.php
```diff
@@ -0,0 +1,393 @@
+<?php
+
+use App\Enums\CommunityConnectorHasLivedExperience;
+use App\Enums\ConsultingService;
+use App\Enums\IndividualRole;
+use App\Enums\MeetingType;
+
+dataset('individualIsPublishable', function () {
```
greatislander, line 8: Nice use of datasets! This is new to me but perfect for this situation.

greatislander (approved):
This looks great— thanks, @jobara!

PR #1153, opened 2022-10-07
feat: browse all projects page

### LINKED ISSUE #306: Browse projects

chosww:
Updates browse all projects page with new livewire component with search and filter functions.

### Prerequisites


greatislander:
@chosww Can you try to increase the coverage for the AllProjects component?

app/Http/Livewire/AllProjects.php
```diff
@@ -0,0 +1,236 @@
...
+    public function updatedStatuses()
+    {
+        if (! is_array($this->statuses)) {
+            return;
+        }
+        $this->statuses = array_filter($this->statuses,
+            function ($status) {
+                return $status != false;
+            }
+        );
```
greatislander: Could this be extracted into a method? Something like:

```php
public function filterItems(array $items) {
  return array_filter($items, function ($item) {
    return $item != false;
  });
}
```

And then you could call `$this->statuses = $this->filterItems($this->statuses);`

greatislander: I'm also curious how `false` would end up in an array? Let's discuss tomorrow, @chosww.

app/Http/Livewire/AllProjects.php
```diff
@@ -0,0 +1,236 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Livewire\Component;
+use Livewire\WithPagination;
+use Spatie\LaravelOptions\Options;
+
+class AllProjects extends Component
+{
+    use WithPagination;
+
+    public string $query = '';
```
greatislander: I think it would be a good idea to change this to `$searchQuery` to differentiate it from all the `$query` calls in the `render()` method.

app/Http/Livewire/AllProjects.php
```diff
@@ -0,0 +1,236 @@
...
+                return $recruitmentMethod != false;
+            }
+        );
+    }
+
+    public function render()
+    {
+        return view('livewire.all-projects', [
+            'projects' => Project::status('published')
+                    ->where(
```
greatislander: I think this should only be applied with a `when` clause only when the search query is present, otherwise there's an extra database query just to search for items which match an empty string.

greatislander (changes requested):
I left some inline comments. Additionally, I think it would be a good idea to use the same approach that's used to show the count of search results to show a count of results matching the applied filters. One could do something like:

`3 projects match your applied filters.`

I also think there should be a "Clear All" button at the end of the list of filters.

Lastly, there's a few tweaks to spacing and layout that we should look at together. But overall this is great work!

resources/views/livewire/all-projects.blade.php
```diff
@@ -22,22 +22,35 @@
...
+            $seekingGroups ||
+            $initiators ||
+            $meetingTypes ||
+            $locations ||
+            $compensations ||
+            $sectors ||
+            $impacts ||
+            $recruitmentMethods)
+            <p class="h4">
+                {{ __(':count projects match your apllied filters', ['count' => $projects->total()]) }}
```
greatislander: Typo: `applied`

greatislander (changes requested):

app/Http/Livewire/AllProjects.php
greatislander: @chosww Just wanted to check about this— I see you created the abstract method but I'm wondering if it's necessary? Can false values end up in these arrays?

chosww: the $item parameter becomes false when I uncheck the checkbox, so I think that this array_filter is using callback value from the property update

greatislander: Oh, interesting. I'll take a look at that.

greatislander (changes requested):
Styling:

- set default top and bottom margins of `ul` elements to 0
- remove `stack` class from `.filters`, add top margin to `.filters`
- give the `[role="alert"]` a minimum height of 1.5rem
- remove vertical gap between `Search` label and input
- add space between search bar and filters/cards

app/Http/Livewire/AllProjects.php
```diff
@@ -0,0 +1,185 @@
...
+                    })
+                    ->orderBy('published_at', 'desc')
+                    ->paginate(20),
+            'statusesData' => [
+                ['value' => 'upcoming', 'label' => __('Upcoming')],
+                ['value' => 'inProgress', 'label' => __('In Progress')],
+                ['value' => 'completed', 'label' => __('Completed')],
+            ],
+            'seekingsData' => [
+                ['value' => 'participants', 'label' => __('Seeking Individual Consultation Participatns')],
```
greatislander: Typo: `Participants`

app/Http/Livewire/AllProjects.php
```diff
@@ -0,0 +1,185 @@
...
+                ['value' => 'connectors', 'label' => __('Seeking Community Connectors')],
+                ['value' => 'organizations', 'label' => __('Seeking Community Organizations to Consult with')],
```
greatislander: `consult` should be lower case

resources/views/livewire/all-projects.blade.php
```diff
@@ -0,0 +1,233 @@
...
+        @clear-flash-message.window="visible = false"
+        @remove-flash-message.window="setTimeout(() => visible = false, 5000)">
+        <div x-show="visible" x-transition:leave.duration.500ms>
+            @if (session()->has('message'))
+                <x-hearth-alert type="success">
+                    {!! Str::markdown(session('message')) !!}
+                </x-hearth-alert>
+            @endif
+        </div>
+    </div>
```
greatislander: I don't think we need this as this component doesn't flash any messages.

resources/views/livewire/all-projects.blade.php
```diff
@@ -0,0 +1,233 @@
...
+        <div class="repel">
+            <x-hearth-input name="searchQuery" type="search" wire:model.defer="searchQuery" wire:search="search" />
+            <button>{{ __('Search') }}</button>
+        </div>
+    </form>
+
+    <div role="alert">
+        @if ($searchQuery)
+            <p class="h4">
+                {{ __(':count results for “:searchQuery', ['count' => $projects->total(), 'searchQuery' => $searchQuery]) }}
```
greatislander: Can you add a period at the end of this sentence?

greatislander: We'll want to use `trans_choice` to handle the plural forms: https://laravel.com/docs/9.x/localization#pluralization

greatislander (changes requested):

greatislander:
One other change @chosww—we need to exclude "Cross-disability" from the disability types list in the AllProjects component.

chosww:
Hey @greatislander, I had to update minimum height to be space-9 to fit in the h4 content without UI getting pushed, and I still have to figure out the trans_choice with our way of localize strings.

resources/views/livewire/all-projects.blade.php
```diff
@@ -0,0 +1,221 @@
...
+        @endif
+    </div>
+
+    <div class="stack with-sidebar with-sidebar:2/3">
+        <div class="filters">
+            <h2 class="visually-hidden">{{ __('Filters') }}</h2>
+            <x-expander :level="3">
+                <fieldset class="filter__options field @error('status') field--error @enderror">
+                    <x-slot name="summary">{{ __('Status') }}</x-slot>
+                    <ul class="mb-4" role="list">
```
greatislander: Let's make this a style specific to filter `<ul>` elements in our stylesheet (maybe make and import a new stylesheet partial in `components/_filters.css` and import it into `app.css`):

```css
.filters .expander .ul {
  margin-block-end: var(--space-4);
}
```

resources/css/layouts/_form.css
```diff
@@ -10,7 +10,7 @@ fieldset > legend + * {
     margin-block-start: var(--space-5);
 }
 
-form .repel:last-child {
+form .repel:not([class*="my-"]):last-child {
```
greatislander: Not sure about this approach? Let's discuss.

resources/css/layouts/_app.css
```diff
@@ -1 +1,9 @@
 /* This file contains all styles related to the application layout. */
+
+ul {
+    margin-block: 0;
+}
+
+.filters {
```
greatislander: This should go in the `components/_filters.css` partial I mentioned.

resources/css/layouts/_app.css
```diff
@@ -1 +1,9 @@
 /* This file contains all styles related to the application layout. */
+
+ul {
+    margin-block: 0;
+}
```
greatislander: I'd rather have this in `base/_base.css`. I think that's the right place for it rather than the app layout.

resources/css/components/_alert.css
```diff
@@ -1,4 +1,8 @@
 /* This file contains all styles related to the alert component. */
+[role="alert"] {
+    min-height: var(--space-9);
```
greatislander: I think it would be better to scope this to the particular use case. Maybe give the `<div role="alert">` a class, `.search-and-filter-results`, and use that class for the style. I don't think this rule should necessarily apply to all alerts.

greatislander (changes requested):

greatislander (approved):

cherylhjli:
- Supporter should be under the disability and deaf group filter
- Can we move the “select none” button up under “[number] projects match your applied filters” and rename it “clear all”? And only show it if there are filters applied? Currently it’s pretty buried and I only know where it is because I’ve seen this page before!
- Checkboxes - if I check or uncheck more than a few, boxes that I’ve already checked/unchecked go into the opposite state. (I can demo this in screenshare)

PR #1175, opened 2022-10-13
feat: join engagements via open calls

greatislander:
- [x] Allows individual participants to join engagements via an open call
- [x] Removes the ability for individual participants to express interest in projects

### Prerequisites


chosww (approved):

cherylhjli:
When I was going through this flow, I signed up for an engagement but said 'no' to sharing my access needs. Then I landed back on the engagement page with a success notification that said my access needs preferences are saved. However, I didn't realize whether I have signed up for the engagement or not, if I said no - despite the fact that we designed these flows together!

I think we make it more obvious to the user that they have successfully signed up, and that this confirm your access needs page is just an additional step. Currently feels like it's a step to get signed up. I think one way to do this is to have in h1 "You have successfully signed up" so it's nice and big, and move "confirm your access needs" into h2. We can talk more about this tomorrow if you need to!

PR #1225, opened 2022-10-24
feat: manage accounts (resolves #1211)

### LINKED ISSUE #1211: Admin: manage individual and (regulated) organization approval/orientation status

Administrators should be able to approve, suspend, and mark individual/organizational/regulated organizational accounts as having completed their orientation.

https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=3715%3A94295

greatislander:
Resolves #1211.

- [x] Approve individuals, organizations, and regulated organizations
- [x] Suspend and unsuspend individuals, organizations, and regulated organizations
- [x] Add approval status to the publishable checks for their pages
- [x] Notify individuals, organizations, and regulated organizations of approval, suspension, and unsuspension
- [x] Restrict access to their pages based on suspension status
- [x] Restrict their activity based on suspension status

### Prerequisites


app/Policies/OrganizationPolicy.php
```diff
@@ -41,11 +53,15 @@ public function view(User $user, Organization $organization): Response
...
+            return ($user->isAdministratorOf($organization) || $user->isAdministrator()) && $organization->isPublishable()
                 ? Response::allow()
                 : Response::denyAsNotFound();
         }
 
+        if ($user->isSuspended() && ($user->isAdministratorOf($organization) || $user->isAdministrator()) && $organization->isPublishable()) {
+            return Response::allow();
+        }
+
         return $user->individual || $user->organization || $user->regulated_organization
```
jobara: Should also allow admin's to view the page here.

app/Http/Livewire/ManageAccounts.php
```diff
@@ -0,0 +1,171 @@
...
+        session()->flash('message', __(':account has been suspended.', ['account' => $individual->name]));
+
+        $this->dispatchBrowserEvent('add-flash-message');
+
+        $this->dispatchBrowserEvent('remove-flash-message');
+    }
+
+    public function suspendAccount(int $id, string $class)
+    {
+        $classname = "App\\Models\\{$class}";
```
jobara: Rather than manually constructing the classname, you could call [`get_class()`](https://www.php.net/manual/en/function.get-class.php). 

Is it possible just to pass in the model instance directly. It looks like you have access to the model when this method is called.

app/Models/Individual.php
```diff
@@ -341,7 +348,7 @@ public function upcomingContractedProjects(): MergedRelation
      *
      * @return bool
      */
-    public function isPublishable(): bool
+    public function isPublishable(bool $ignoreSuspension = false): bool
```
jobara: I'm guessing this parameter isn't necessary. Similar for other cases it is added.

app/Models/Individual.php
```diff
@@ -378,6 +385,10 @@ public function isPublishable(): bool
                 return false;
             }
 
+            if (! $this->user->checkStatus('approved')) {
+                return false;
```
jobara: As [mentioned](https://matrix.to/#/!cTDzVKxAmJRmfbSdHL:matrix.org/$W_WnUZR_11mHhffIbwS2Blq5f9F2N4S8IpO-_Alcr2c?via=matrix.org) in the channel. This makes it unpreviewable as well.

app/Models/Individual.php
greatislander: Yes, I removed them.

app/Http/Livewire/ManageAccounts.php
greatislander: I called `get_class()` in the view initially and ran into an issue where the backslashes were being escaped— it's probably easier to do that and handle the escaped values though. I haven't had any luck passing the model instance directly, but I can look into that again.

app/Notifications/AccountSuspended.php
```diff
@@ -0,0 +1,56 @@
...
+        ];
+    }
+
+    public function getCapabilities($notifiable): string
+    {
+        if ($notifiable instanceof Individual && ! $notifiable->isConnector() && ! $notifiable->isConsultant()) {
+            return __('You will not be able to edit any information in your account.');
+        }
+
+        return __('You will not be able to edit any information in your account. If you published your page already, it has been unpublished.');
```
jobara: I guess this is true, but now I'm wondering if it should actually be unpublished (that is change the published state in the model). However, that would require them to re-publish the page when they are unsuspended.

Assuming we keep things as is, what would the expectation be for the admin when viewing a suspended user's page. It isn't published but it also isn't draft; which may have an effect on what we do for #1232

app/Policies/DefinedTermPolicy.php
```diff
@@ -10,61 +10,28 @@ class DefinedTermPolicy
...
+//
+//    public function update(User $user, DefinedTerm $definedTerm): bool
+//    {
+//        return false;
+//    }
+//
+//    public function delete(User $user, DefinedTerm $definedTerm): bool
+//    {
+//        return false;
+//    }
```
jobara: Should these commented lines be removed?

app/Policies/EngagementPolicy.php
```diff
@@ -6,18 +6,34 @@
...
+        if ($user->isSuspended() && $user->isAdministratorOf($engagement->project->projectable) && $engagement->isPublishable()) {
+            return Response::allow();
+        }
+
+        if ($user->isAdministrator() && $engagement->isPublishable()) {
+            return Response::allow();
+        }
+
         return
             ($user->individual || $user->organization || $user->regulated_organization)
```
jobara: probably needs to allow admin users here. Can this pass for all authorized users, or is there a need to check all these different types?

app/Policies/OrganizationPolicy.php
```diff
@@ -6,14 +6,26 @@
...
 {
     use HandlesAuthorization;
 
-    public function before(User $user): null|bool
+    public function before(User $user, string $ability): null|Response
     {
-        return $user->isAdministrator() ? true : null;
+        if ($user->isAdministrator() && $ability === 'viewAny') {
+            return Response::allow();
+        }
```
jobara: As this only appears to be needed for the `viewAny` ability, I think it could be moved there.

app/Policies/ProjectPolicy.php
```diff
@@ -6,14 +6,26 @@
...
 {
     use HandlesAuthorization;
 
-    public function before(User $user): null|bool
+    public function before(User $user, string $ability): null|Response
     {
-        return $user->isAdministrator() ? true : null;
+        if ($user->isAdministrator() && $ability === 'viewAny') {
+            return Response::allow();
+        }
```
jobara: Can be moved to the viewAny function

app/Policies/ProjectPolicy.php
```diff
@@ -25,44 +37,32 @@ public function viewAny(User $user): Response
...
+        if ($user->isSuspended() && $user->isAdministratorOf($project->projectable) && $project->isPublishable()) {
+            return Response::allow();
+        }
+
+        if ($user->isAdministrator() && $project->isPublishable()) {
+            return Response::allow();
+        }
+
         return
             $user->individual || $user->organization || $user->regulated_organization
```
jobara: should admin users have access to this page too?

app/Policies/RegulatedOrganizationPolicy.php
```diff
@@ -41,11 +53,15 @@ public function view(User $user, RegulatedOrganization $regulatedOrganization):
...
+            return ($user->isAdministratorOf($regulatedOrganization) || $user->isAdministrator()) && $regulatedOrganization->isPublishable()
                 ? Response::allow()
                 : Response::denyAsNotFound();
         }
 
+        if ($user->isSuspended() && ($user->isAdministratorOf($regulatedOrganization) || $user->isAdministrator()) && $regulatedOrganization->isPublishable()) {
+            return Response::allow();
+        }
+
         return $user->individual || $user->organization || $user->regulated_organization
```
jobara: Should the admin have access? Same for other cases too.

app/Notifications/AccountSuspended.php
greatislander: @cherylhjli Can we adjust this copy to reflect the fact that pages remain published but are simply hidden from other users— something like "Your page will no longer be shown to other users of this website."

jobara:
@cherylhjli, @greatislander and I were talking today about the need for a banner/message to be displayed to admins when they are viewing content that is for a suspended user. There may need to be copy and/or other design considerations for this.

app/Http/Livewire/ManageAccounts.php
greatislander: Resolved in e634289e.

app/Notifications/AccountSuspended.php
cherylhjli: that sounds good @greatislander

cherylhjli:
@jobara @greatislander I think we can use this design here: https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=5101%3A107877

Copy can be something like, "This account has been suspended." Thoughts?

PR #1284, opened 2022-11-09
feat: display sign language videos in content (Resolves #1268)

### LINKED ISSUE #1268: Display sign language videos in content

jobara:
Resolves #1268 

Adds a component and model for displaying Sign Language videos along side site content. Placeholders have been added to a few pages, including the home page and the about individual page.

To test:

- Login and enable the sign language translations setting from the web accessibility page
- Visit the home page and the about individual page to create the models
- Modify the database by adding in the video URLs. (you could also run the Interpretation factory and set the relevant, route, name, and namespaces).

```json
{
	"ase": "https://vimeo.com/766454375",
	"fcs": "https://vimeo.com/766455246"
}
```

- go back to the home page and about individual page
- click on the "sign language" buttons to see the video displayed.
- if you change between French and English you'll get different videos.

### Prerequisites


greatislander (commented):
This looks great! I have a couple of questions more than requested changes.

- The namespace is used if provided, otherwise we fall back to the route, correct?
- I know we aren't providing an interface for this yet but I ran into an issue where I created an interpretation, then visited the page, and a duplicate was created. I think perhaps it wasn't an exact match because I had provided videos but I'm not sure. Then my videos weren't loading because the first one (created when I visited the page) was used.
- Is it possible to reset the Vimeo player to the start when the video is collapsed, and start it again when it's re-opened? Maybe it's okay as is. I'm just imagining someone may not want to have to restart manually if they've hidden and then shown the video again.

jobara:
> * The namespace is used if provided, otherwise we fall back to the route, correct?

Yes, if a namespace is explicitly provided it uses that, otherwise it uses the route for the namespace. We have separate route attribute in the model so that we can create a link back, if possible, to the originating location. This would be a feature for the admin page though, so it isn't used yet.

> * I know we aren't providing an interface for this yet but I ran into an issue where I created an interpretation, then visited the page, and a duplicate was created. I think perhaps it wasn't an exact match because I had provided videos but I'm not sure. Then my videos weren't loading because the first one (created when I visited the page) was used.

I wonder if this is related to the above comment.

> * Is it possible to reset the Vimeo player to the start when the video is collapsed, and start it again when it's re-opened? Maybe it's okay as is. I'm just imagining someone may not want to have to restart manually if they've hidden and then shown the video again.

I've switched to using Alpine and using Vimeo's JS methods to create the player. It should now start when expanded, stop when collapsed, and after reaching the end of the video playback position is reset to the start.

database/migrations/2022_11_08_181911_change_type_user_sign_langague_translations.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
greatislander: Not really a functional issue but can you fix the typo in this filename?

greatislander (changes requested):
@jobara Looks great— can you rename that migration? Then I'll merge.

database/migrations/2022_11_08_181911_change_type_user_sign_langague_translations.php
jobara: Good catch. I've pushed up a fix.

greatislander (approved):

PR #1302, opened 2022-11-15
feat: add sign language to language menu (resolves #1256)

### LINKED ISSUE #1256: Add Sign Language options to Language drop down menu
The language drop down provides options for switching between English and French text. There should also be an option(s) to enable/disable the sign language (ASL, LSQ) videos. 

This will allow those without an account to enable the sign language videos feature.

jobara:
Resolves #1256 

Adds the sign language options to the Language Menu, and hooks this up to the users settings when logged. New users will pick up the setting from the cookie if available, and the cookie will be removed when the user logs out.

The interpretations component has also been updated to create/fetch instances when the sign_language_translations cookie is set to true.

### Prerequisites


jobara:
@cherylhjli and @greatislander, one thing to note about this implementation is that if someone were to open the link, from the language menu, in a new tab, it wouldn't trigger the change to the sign language setting. It would only change the written language. So for example if you were on English and opening the ASL link in a new tab, it would still be in English. If you opened the LSQ link in a new tab, it would be in French but without the LSQ.

If this is an issue, potentially we could add query parameter to the link in the language menu, check the presence of that query parameter via middleware to perform all the necessary operations before passing it along to the regular controller. This would have to be added for every route though, and would introduce a bit of overhead for all requests.

Of course there are probably other solutions too, and I'm open to suggestions.

resources/views/components/language-switcher.blade.php
```diff
@@ -0,0 +1,53 @@
...
+{{--
+
+                <x-nav-link :href="localized_route('logout')" x-on:click.prevent="$refs.form.submit()">
+                    {{ __('Sign out') }}
+                </x-nav-link>
+                <form method="POST" action="{{ localized_route('logout') }}" x-ref="form">
+                    @csrf
+                </form>
+
+ --}}
```
greatislander: Looks like this was left here by mistake.

greatislander (changes requested):
One small change before this is merge-ready. I think we should address the opening links in new tab issue separately for the logout link as well.

resources/views/components/language-switcher.blade.php
greatislander:
```suggestion
```

greatislander (approved):

PR #1314, opened 2022-11-16
fix: date picker improvements

### LINKED ISSUE #1311: Date picker bugs
**Prerequisites**

**Describe the bug**
- The “day” is pre-populated with "undefined". If I click into it, it doesn’t disappear and I can’t type anything. (However, if I tab into it and it selects the “undefined”, I can edit it). The only way to get rid of it is to click/focus in, and then click out - then it disappears

Screen Shot 2022-11-15 at 3 58 16 PM

greatislander:
- if value is null, set y/m/d to empty strings
- remove minimum year attributes

Hasn't addressed #1311 as I can't reproduce, but if I can figure out what's happening there I will add the changes to this PR.

### Prerequisites

### DIFF
--- a/resources/js/datePicker.js
+++ b/resources/js/datePicker.js
@@ -8,9 +8,9 @@ export default (initial = "") => ({
     init() {
         this.date = initial;
         let [y, m, d] = this.date.split("-");
-        this.year = y;
-        this.month = m;
-        this.day = d;
+        this.year = y ?? "";
+        this.month = m ?? "";
+        this.day = d ?? "";
     },
     getDate() {
         if (this.year || (this.year !== "" && this.month) || (this.month !== "" && this.day) || this.day !== "") {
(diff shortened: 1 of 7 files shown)


resources/js/datePicker.js
```diff
@@ -8,9 +8,9 @@ export default (initial = "") => ({
     init() {
         this.date = initial;
         let [y, m, d] = this.date.split("-");
-        this.year = y;
-        this.month = m;
-        this.day = d;
+        this.year = y ?? "";
+        this.month = m ?? "";
+        this.day = d ?? "";
```
jobara: FOr the destructing assignment, I think you can actually set a default value. Something like:

```javascript
let [y = "", m = "", d = ""] = this.date.split("-");
```

and if so, I wonder if you could then actually do:

```javascript
[this.year = "", this.month = "", this.day =""] = this.date.split("");
```

It isn't a big deal either way though, so what is here is fine too.

greatislander: Oh I'll try that— good call, @jobara.

greatislander: You can't do the second one (destructuring to `this.year` etc). But your first example works!

jobara (approved):

PR #1326, opened 2022-11-17
fix: update copy on access needs settings and confirmation (fix #1298, fix #1324)

### LINKED ISSUE #1298: Participant sign up for engagements: Confirm access needs
**Prerequisites**

**Describe the bug**
When participants sign up for access needs, change the access need page copy to appear like so:
https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=3973%3A95155&t=LbqixbxDHS1z5bCk-1

We've decided that we're not going to use the word "anonymous" anywhere because it's not true - even though access needs are in a separate tab as name on the Manage Participants page for FRO's, there's still a chance FRO's can figure out who has what access needs if they monitor who signs up. This copy adjustment changes the expectation for participants - it doesn't promise any access needs will be shared anonymously.

### LINKED ISSUE #1324: Access need: Change copy under h1
**Prerequisites**

**Describe the bug**
Previous copy said that access needs would be kept anonymous. That is not true, as we cannot guarantee anonymity. Please change copy right underneath H1 "Access Needs" to this:

"When you sign up to participate in consultations, the access needs you check off below will be shared with the organization you are working with so they can meet them. 

Most of the time, we will not need to tell them who requested the access need. However, for some of the access needs, we will - and we’ll tell you which ones below."

greatislander:
- [x] Resolves #1298
- [x] Resolves #1324

### Prerequisites


resources/views/engagements/confirm-access-needs.blade.php
```diff
@@ -1,26 +1,34 @@
...
+                    'access_needs_settings_link' =>
+                        '<a href="' .
+                        localized_route('settings.edit-access-needs') .
+                        '">' .
+                        __('access needs settings') .
+                        '</a>',
+                    'projectable' => $project->projectable->name,
+                ],
             ),
         ) !!}
```
jobara: This is using `Str:markdown` but the string doesn't contain markdown; however there is html added. I think we can either remove `Str:markdown` or add the links with markdown syntax.

jobara (changes requested):

resources/views/engagements/confirm-access-needs.blade.php
greatislander: Good call. It used to have Markdown. I think you are right that we can remove it and wrap in a paragraph.

jobara (approved):
The code changes appear good. @cherylhjli could you QA to make sure that everything is correct in terms of language and presentation.

PR #1329, opened 2022-11-17
feat: simplify language selection, improve sign out experience

### LINKED ISSUE #1305: Opening the Sign Out link in a new page throws an error
**Prerequisites**

**Describe the bug**

If you try to open the Sign Out link in a new tab or window it throws the error:

```error
The GET method is not supported for this route. Supported methods: POST.
```

**To Reproduce**
1. Sign in to any account
2. Open the Sign out link in a new tab or window (e.g right click Sign Out and select open in new tab)
3. Notice the error

**Expected behavior**

I'd expect that a page opens without error. However, what page is less clear. Some potential options:

- A "logout" page that will confirm you are logged out, and internally (in the controller) performs the logout operation.
- A redirect from a logout end point that will first perform the logout operation and then redirect to the login page.
- etc.

greatislander:
- [x] Adds `asl` and `lsq` to site locales, simplifying language selection
- [x] Remove sign language preference from signup and language settings (sign languages can now be selected directly)
- [x] Remove sign language preference from website accessibility settings
- [x] Update language changer
- [x] Update translatable inputs and textareas
- [x] Change sign out to button + form submission (resolve #1305)

### Prerequisites


app/Actions/Fortify/CreateNewUser.php
```diff
@@ -65,9 +63,7 @@ public function create(array $input): User
                     Rule::notIn([UserContext::Administrator->value]),
                 ],
                 'extra_attributes' => 'nullable|array',
-                'locale' => ['required', Rule::in(config('locales.supported', ['en', 'fr']))],
-                'signed_language' => 'nullable|string|in:ase,fcs',
-                'sign_language_translations' => 'nullable|boolean',
+                'locale' => ['required', Rule::in(config('locales.supported', ['en', 'fr', 'asl', 'lsq']))],
```
jobara: rather than hardcoding the locales here, you can probably source them from the config. I think `config('locales.supported')` would work.

app/Http/Requests/UpdateLanguagePreferencesRequest.php
```diff
@@ -17,9 +17,8 @@ public function rules(): array
         return [
             'locale' => [
                 'required',
-                Rule::in(config('locales.supported', ['en', 'fr'])),
+                Rule::in(config('locales.supported', ['en', 'fr', 'asl', 'lsq'])),
```
jobara: Similar to above, can these be sourced from the config?

app/Models/Organization.php
```diff
@@ -130,9 +130,13 @@ protected static function booted()
 
     public function getSlugOptions(): SlugOptions
     {
-        return SlugOptions::createWithLocales(['en', 'fr'])
+        return SlugOptions::createWithLocales(['en', 'fr', 'asl', 'lsq'])
```
jobara: Source locales from config.

app/Models/RegulatedOrganization.php
```diff
@@ -127,9 +127,16 @@ protected static function booted()
 
     public function getSlugOptions(): SlugOptions
     {
-        return SlugOptions::createWithLocales(['en', 'fr'])
+        return SlugOptions::createWithLocales(['en', 'fr', 'asl', 'lsq'])
```
jobara: Source locales from config.

app/Models/RegulatedOrganization.php
```diff
@@ -127,9 +127,16 @@ protected static function booted()
...
             ->generateSlugsFrom(function (RegulatedOrganization $model, $locale): string {
-                return $model->getTranslation('name', $locale);
+                if (in_array($locale, ['fr', 'lsq'])) {
+                    ray($model->getTranslation('name', 'fr'));
```
jobara: Is this just for debugging? Should it be removed?

app/Models/RegulatedOrganization.php
```diff
@@ -127,9 +127,16 @@ protected static function booted()
...
+
+                    return $model->getTranslation('name', 'fr');
+                }
+                ray($model->getTranslation('name', 'en'));
```
jobara: Similar to above, is this just for debugging?

app/Models/Resource.php
```diff
@@ -59,9 +59,13 @@ class Resource extends Model
      */
     public function getSlugOptions(): SlugOptions
     {
-        return SlugOptions::createWithLocales(['en', 'fr'])
+        return SlugOptions::createWithLocales(['en', 'fr', 'asl', 'lsq'])
```
jobara: Source locales from config.

app/Models/Resource.php
```diff
@@ -59,9 +59,13 @@ class Resource extends Model
...
             ->generateSlugsFrom(function (Resource $model, $locale): string {
-                return $model->getTranslation('title', $locale);
+                if (in_array($locale, ['fr', 'lsq'])) {
+                    return $model->getTranslation('title', 'fr');
+                }
+
+                return $model->getTranslation('title', 'en');
```
jobara: This is essentially the third time this function was written. I think it would make sense to move this to a helper or trait that can be applied to the models. I think the only difference is the need to parameterize the attribute name passed to `getTranslation`.

app/Models/ResourceCollection.php
```diff
@@ -43,9 +43,13 @@ class ResourceCollection extends Model
      */
     public function getSlugOptions(): SlugOptions
     {
-        return SlugOptions::createWithLocales(['en', 'fr'])
+        return SlugOptions::createWithLocales(['en', 'fr', 'asl', 'lsq'])
```
jobara: Source locales from config.

config/locales.php
```diff
@@ -25,7 +27,7 @@
     */
 
     'paired_sign_language' => [
-        'en' => 'ase',
-        'fr' => 'fcs',
+        'en' => 'asl',
+        'fr' => 'lsq',
     ],
```
jobara: I don't think this is used anymore and can be removed.

app/View/Components/TranslationPicker.php
```diff
@@ -14,7 +14,7 @@ class TranslationPicker extends Component
 
     public function __construct(?array $languages = null, ?array $availableLanguages = null)
     {
-        $this->languages = $languages ?? ['en', 'fr', 'ase', 'fcs'];
+        $this->languages = $languages ?? ['en', 'fr', 'asl', 'lsq'];
```
jobara: Source locales from config.

app/helpers.php
```diff
@@ -89,13 +97,13 @@ function ($language) {
...
-        $fcs = $result['fcs'];
-        unset($result['fcs']);
+        $asl = $result['asl'];
+        unset($result['asl']);
+        $lsq = $result['lsq'];
+        unset($result['lsq']);
 
         if ($signed) {
-            foreach (['fcs', 'ase', 'fr', 'en'] as $code) {
+            foreach (['lsq', 'asl', 'fr', 'en'] as $code) {
```
jobara: Consider sourcing the locales from config. In this case you might want to provide a sort order. However, in this case just reversing the array is probably enough.

```suggestion
            foreach (array_reverse(config('locales.supported')) as $code) {
```

app/helpers.php
```diff
@@ -170,17 +178,21 @@ function get_signed_language_for_written_language(string $code): string
      */
     function get_language_exonym(string $code, string $locale = '', bool $capitalize = true): null|string
     {
-        if ($locale === '') {
-            $locale = locale();
-        }
+        $locale = $locale === '' ? locale() : $locale;
```
jobara: Minor but might be easier to make $locale nullable and default to `null`. Then here you could do the follow:

```suggestion
        $locale ??= locale();
```

Or if you can use `locale()` as the default in the parameters you can skip this line entirely.

app/helpers.php
```diff
@@ -170,17 +178,21 @@ function get_signed_language_for_written_language(string $code): string
...
+        $locale = match ($locale) {
+            'asl' => 'en',
+            'lsq' => 'fr',
+            default => $locale
+        };
 
         switch ($code) {
-            case 'fcs':
+            case 'lsq':
                 return trans('locales.'.$code, [], $locale);
```
jobara: `lsq` doesn't support the `capitalized` option. The localization currently has it capitalized by default. So you could probably do the reverse and lowercase this if the capitalize argument is false. Although it would be more robust to handle both lower and uppsercasing so we aren't dependent on how it is written in the translation file.

database/factories/EngagementFactory.php
```diff
@@ -13,7 +13,7 @@ public function definition(): array
             'published_at' => date('Y-m-d h:i:s', time()),
             'project_id' => Project::factory(),
             'name' => ['en' => 'Workshop'],
-            'languages' => ['en', 'fr', 'ase', 'fcs'],
+            'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: Source locales from config.

database/factories/OrganizationFactory.php
```diff
@@ -11,7 +11,7 @@ public function definition(): array
         return [
             'name' => $this->faker->company(),
             'type' => $this->faker->randomElement(['representative', 'support', 'civil-society']),
-            'languages' => ['en', 'fr', 'ase', 'fcs'],
+            'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: source locales from config.

database/factories/ProjectFactory.php
```diff
@@ -30,7 +30,7 @@ public function definition()
             'projectable_type' => 'App\Models\RegulatedOrganization',
             'projectable_id' => RegulatedOrganization::factory(),
             'name' => ['en' => 'My accessibility project – '.Carbon::parse($start)->format('F Y')],
-            'languages' => ['en', 'fr', 'ase', 'fcs'],
+            'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: source locales from config.

database/factories/RegulatedOrganizationFactory.php
```diff
@@ -24,7 +24,7 @@ public function definition(): array
         return [
             'name' => $this->faker->company(),
             'type' => $this->faker->randomElement(['government', 'business', 'public-sector']),
-            'languages' => ['en', 'fr', 'ase', 'fcs'],
+            'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: source locales from config

database/seeders/TestDataSeeder.php
```diff
@@ -74,7 +74,7 @@ public function run(): void
                 'email' => 'k@accessibilityexchange.ca',
                 'preferred_contact_person' => 'me',
                 'preferred_contact_method' => 'email',
-                'constituentLanguages' => ['ase', 'fcs', 'en', 'fr'],
+                'constituentLanguages' => ['asl', 'lsq', 'en', 'fr'],
```
jobara, line 77: Source locales from config.

database/seeders/TestDataSeeder.php
```diff
@@ -589,7 +589,7 @@ public function run(): void
         $engagementsForTesting = [
             [
                 'name' => ['en' => 'Workshop'],
-                'languages' => ['en', 'fr', 'ase', 'fcs'],
+                'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: Source locales from config.

database/seeders/TestDataSeeder.php
```diff
@@ -602,7 +602,7 @@ public function run(): void
             ],
             [
                 'name' => ['en' => 'Focus Group'],
-                'languages' => ['en', 'fr', 'ase', 'fcs'],
+                'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: Source locales from config.

database/seeders/TestDataSeeder.php
```diff
@@ -616,7 +616,7 @@ public function run(): void
             ],
             [
                 'name' => ['en' => 'Expert Analysis'],
-                'languages' => ['en', 'fr', 'ase', 'fcs'],
+                'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: Source locales from config.

resources/views/components/interpretation.blade.php
```diff
@@ -1,20 +1,23 @@
...
-            portrait: false,
-            responsive: true,
-            speed: true,
-            title: false
-        })" x-init="$watch('expanded', value => togglePlayback(value))"
-            @ended="player().setCurrentTime(0)" x-show="expanded" x-cloak>
-        </div>
-    @endif
-</div>
+@if (in_array(locale(), ['asl', 'lsq']))
```
jobara: The  `is_signed_language` helper function should be used here instead.

tests/Datasets/EngagementIsPublishable.php
```diff
@@ -3,7 +3,7 @@
 dataset('engagementIsPublishable', function () {
     $baseModel = [
         'name' => ['en' => 'Workshop'],
-        'languages' => ['en', 'fr', 'ase', 'fcs'],
+        'languages' => ['en', 'fr', 'asl', 'lsq'],
```
jobara: Source locales from config.

jobara (changes requested):

jobara:
@greatislander on the pages, the `<html>`'s `lang` attribute should probably only be `en` or `fr` and not `lsq` or `asl`. Same goes for the `hreflang` attributes, or maybe in those cases use `ace` and `fcs`.

jobara:
@greatislander also want to confirm that the `signed_language_for_interpretation` in [UpdateAccessNeedsRequest.php](https://github.com/accessibility-exchange/platform/blob/192be4201b9f278bf524569c1685f7c8a00f8113/app/Http/Requests/UpdateAccessNeedsRequest.php#L31) should still be using `ace` and `fcs`.

tests/Datasets/EngagementIsPublishable.php
greatislander: The `config` helper is not available within `dataset`.

greatislander:
> @greatislander also want to confirm that the `signed_language_for_interpretation` in [UpdateAccessNeedsRequest.php](https://github.com/accessibility-exchange/platform/blob/192be4201b9f278bf524569c1685f7c8a00f8113/app/Http/Requests/UpdateAccessNeedsRequest.php#L31) should still be using `ace` and `fcs`.

Resolved in 91643a97.

greatislander:
> @greatislander on the pages, the `<html>`'s `lang` attribute should probably only be `en` or `fr` and not `lsq` or `asl`. Same goes for the `hreflang` attributes, or maybe in those cases use `ace` and `fcs`.

Resolved in 798d04ff.

app/Providers/FortifyServiceProvider.php
```diff
@@ -82,8 +82,8 @@ public function boot()
         Fortify::registerView(function () {
             return view('auth.register', [
                 'signLanguages' => Options::forArray([
-                    'ase' => get_language_exonym('ase', 'fr'),
-                    'fcs' => get_language_exonym('fcs', 'fr'),
+                    'asl' => get_language_exonym('asl', 'fr'),
```
jobara: Should the locale argument be 'en'?

resources/views/components/interpretation.blade.php
```diff
@@ -1,20 +1,23 @@
...
-            title: false
-        })" x-init="$watch('expanded', value => togglePlayback(value))"
-            @ended="player().setCurrentTime(0)" x-show="expanded" x-cloak>
-        </div>
-    @endif
-</div>
+@if (is_signed_language(locale()))
+    <div id="{{ $id }}" {{ $attributes->class(['interpretation', 'expander', 'stack']) }}
+        x-data="{ expanded: false }">
+        @if ($videoSrc)
```
jobara: Should probably move the `$videoSrc` check into the first conditional, that way the parent `<div>` isn't rendered when there is no video provided but the sign language locale is set.

jobara: Although doing so would mean that the links from the admin panel won't work if the sign language locale isn't set. I suppose we could force those URLs into one of the appropriate sign language locale. Not making the above change will cause a bit of a shift in the layout for the empty div.

app/Providers/FortifyServiceProvider.php
jobara: I'm getting an error when trying to create a new user with the language set to ASL. 

app / Providers / FortifyServiceProvider.php : 93

```error
Spatie\LaravelOptions\SelectOption::__construct(): Argument #1 ($label) must be of type string, array given, called in /Users/jo/Documents/git/accessibility-exchange/platform/vendor/spatie/laravel-options/src/Providers/EnumProvider.php on line 36
```

greatislander: Resolved in 4869635d.

resources/views/components/interpretation.blade.php
greatislander: Resolved in https://github.com/accessibility-exchange/platform/pull/1329/commits/c6b0d60c1046d3e4c2eca51b629ffcdcf20080a1 and ffa73f86.

jobara (approved):

PR #1374, opened 2022-12-03
fix(deploy): changed ordering of artisan commands so that storage is …

### LINKED ISSUE #1357: Deployment show "@spaceless", missing dependencies
**Prerequisites**

**Describe the bug**

The deployments appear to be missing the hedronium/spaceless-blade dependency, and as such are showing the raw text `@spaceless` instead of rendering the blade directive from that package. Possibly due to not running `composer install` or it not running properly in the deploy scripts.

**To Reproduce**
1. Go to [staging](https://staging.accessibilityexchange.ca/en)
2. See: `@spaceless`

**Expected behavior**

The blade directive should render it's contents without whitespaces, and not render the name of the directive `@spaceless`

**Screenshots**
@spaceless blade directive name being rendered

marvinroman:
In local testing I found that the ordering of the artisan commands mattered. 
I changed so that the local command which includes `storage:link` should be run first so that optimizations and caching calls all happen in the correct directory.

The revision is to call `php artisan deploy:local` before `php artisan deploy:global`.


.deploy/accessibility-app/entrypoint.sh
```diff
@@ -23,6 +23,8 @@ ln -s $CACHE_PATH /app/bootstrap/cache
 ## fix permissions after syncing to existing storage and cache https://github.com/accessibility-exchange/platform/issues/1236
 chown -R www-data:www-data $FILES_PATH $CACHE_PATH
 
+php artisan deploy:local
+
```
jobara: Does the local deploy need to be guarded by the deploy lock? Are you able to explain a bit about what the deploy lock is preventing?

jobara (commented):
Could you review the contents of the global and local deploys? For example should the caching related tasks be part of the the local deploy, or could they be moved to the global one? Is the storage link the only thing necessary in the local deploy?

.deploy/accessibility-app/entrypoint.sh
marvinroman: The deploy lock is for when more than one container is spinning up at once, it prevents multiple containers from calling the artisan commands within global command. `optimize:clear, event:cache, optimize, migrate`. 

Local doesn't need the lock because they are container specific and it is the main difference between the 2 different commands. The commands it calls are `storage:link, google-fonts:fetch, icons:clear, icons:cache`.

marvinroman: As far as `google-fonts:fetch, icons:clear, icons:cache` I haven't dug in deep on these. They were determined by Ned that they should be done in the local one.
It all depends on the directories that they operate in. If they operate in `/app/storage` or `/app/bootstrap/cache` then I don't see why we wouldn't move them to global. I don't know if it's been tested that way.

marvinroman: Do you want me to research each of these and test them?

greatislander: @marvinroman That would be fantastic. The Google Fonts command comes from https://github.com/spatie/laravel-google-fonts, and the others come from https://github.com/blade-ui-kit/blade-icons.

jobara: @marvinroman maybe we should split off the `google-fonts:fetch, icons:clear, icons:cache` portion into a separate PR if the change is needed. In that way, we can merge the other changes here in for our upcoming release.

marvinroman: @jobara I looked into fonts and icons and they are both stored in permanent storage so I recommend moving to global.

I have included those changes in this PR along with a different locking mechanism that should be more robust and I tested locally through launching 4 replicas at once and seeing global only being logged on a single one. The following command gets an exclusive lock on the `/opt/data/cache` directory which is shared across containers. It is non-blocking meaning it fails immediately if a lock exists and doesn't run the command. `-E 0` is exit code 0 which means successful since it's not an error.
```bash
flock -n -E 0 /opt/data/cache -c "php artisan deploy:global"
``` 

I have also moved the changing of file ownerships to after the caches are reset to make sure that file ownership in these directories are correct since those commands are run as root.
New global commands ordering.
```php
        $this->call('optimize:clear');
        $this->call('icons:clear');
        $this->call('event:cache');
        $this->call('icons:cache');
        $this->call('google-fonts:fetch');
        $this->call('optimize');
        $this->call('migrate', ['--step' => true, '--force' => true]);
```

greatislander (changes requested):
There still appears to be a merge conflict.

marvinroman:
> There still appears to be a merge conflict.

@greatislander @jobara I missed the merge conflict. It's been resolved.

greatislander (approved):

jobara (approved):

PR #1390, opened 2022-12-07
feat: training modules views and controllers (resolves #396, #398, #399, #1399)

### LINKED ISSUE #396: Training modules: view

### LINKED ISSUE #398: Training modules: quiz completion

### LINKED ISSUE #399: Training modules: certificate of completion

### LINKED ISSUE #1399: Enable training user context
This will allow users to register independently in order to complete training courses.

Currently disabled for signup here: https://github.com/accessibility-exchange/platform/blob/0bc99859547a5bb84d9ad87083f43f6571cbabe6/app/Providers/FortifyServiceProvider.php#L85

chosww:
Add views and controllers for training modules (#396).
Add quiz completion (#398). 
Add certificate of completion (#399).
Enable training user context (#1399).

Closes #396, #398, #399, and #1399.

### Prerequisites


app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+use Illuminate\Support\Facades\Auth;
+
+class CourseController extends Controller
+{
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\View\View
+     */
+    public function index(): View
```
greatislander: You can write a test to check that this route is accessible which will increase coverage for this file.

greatislander (changes requested):

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+use Illuminate\Contracts\View\View;
+use Illuminate\Support\Facades\Auth;
+
+class CourseController extends Controller
+{
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\View\View
+     */
```
greatislander: We don't need this doc block.

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+            'courses' => Course::all(),
+        ]);
+    }
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  \App\Models\Course  $course
+     * @return \Illuminate\View\View
+     */
```
greatislander: Doc block not needed.

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+    public function show(Course $course)
```
greatislander: Add a return type hint, e.g.

```suggestion
    public function show(Course $course): View
```

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $user = Auth::user();
+        $courseUser = $user->courses->where('id', $course->id)->first();
```
greatislander: The pivot can be retrieved using `$user->courses->find($course->id)->pivot`.

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+        $finishedCourse = $courseUser->pivot->finished_at ?? null;
```
greatislander: If you set a `$pivot` variable as per the previous comment, you can use the following to get the same result here:

```php
$finishedCourse = $pivot?->finished_at;
```

See: https://stitcher.io/blog/php-8-nullsafe-operator

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,43 @@
...
+        $receivedCertificate = $courseUser->pivot->received_certificate_at ?? null;
+
+        return view('courses.show', [
+            'user' => $user,
+            'course' => $course,
+            'modules' => $course->modules,
+            'finishedCourse' => $finishedCourse && ! $receivedCertificate,
```
greatislander: I think you should pass `$finishedCourse` and `$receivedCertificate` separately. See comments on the view.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,66 @@
...
+            @endforeach
+        </div>
+        @if ($finishedCourse)
+            <form class="stack" action="{{ localized_route('quizzes.show', $course->quiz) }}" novalidate>
+                @csrf
+                <button>{{ __('Take Quiz') }}</button>
+            </form>
+        @else
+            <button type="button" @ariaDisabled>{{ __('Take Quiz') }}</button>
+        @endif
```
greatislander: This should be restructured so that the button is shown but disabled if the course has been finished but the button is not shown if the certificate has been received. You'll need both `$finishedCourse` and `$receivedCertificate` to do so.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,66 @@
...
+            </div>
+        </div>
+    </div>
+    <div class="stack ml-2 mr-2">
+        <h2>{{ __('About this course') }}</h2>
+        @if ($course->video)
+            <video class="w-full" controls>
+                <source src="" type="video/mp4">
+            </video>
+        @endif
```
greatislander: This will be a Vimeo player, so should be handled accordingly.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,66 @@
...
+        <p>{{ $course->introduction }}</p>
+    </div>
+    <div class="stack ml-2 mr-2">
+        <h2>{{ __('Modules') }}</h2>
+        <div class="grid">
+            @foreach ($modules as $module)
+                <div class="flex flex-col">
+                    <video class="w-full" controls>
+                    </video>
```
greatislander: This should be removed.

greatislander (changes requested):
Some initial comments. Will add more in the morning!

app/Http/Controllers/CourseController.php
greatislander: This currently breaks if a user isn't logged in. I think the issue here is more that we shouldn't allow guests to access courses. We'll need middleware on these routes accordingly.

resources/views/emails/quiz/results.blade.php
```diff
@@ -0,0 +1,9 @@
+<x-mail::message>
+    # {{ __('Quiz results') }}
+
+    {{ __('Congraturations') . ', ' . $name . '!' }}
+    {{ __('You have successfully completed course') . ' ' . $course . ' ' . __('and passed the quiz') . '.' }}
+
+    Thanks,
+    {{ config('app.name') }}
```
greatislander: These should not be indented.

app/Models/Question.php
```diff
@@ -36,4 +43,9 @@ public function choices(): HasMany
     {
         return $this->hasMany(Choice::class);
     }
+
+    public function getChoices()
+    {
+        return Options::forArray($this->choices()->pluck('value'), useLabelsAsValue: true)->toArray();
+    }
```
greatislander: Instead of this, choices can be fetched using:

```php
Options::forModels(
	Choice::query()->whereIn('id', $question->choices->pluck('id')->toArray()),
	label: fn(Choice $choice) => $choice->getTranslation('label', 'en')
)->toArray();
```

The `HasTranslations` trait will need to be added to the `Choice` model and the `value` column on the `Choice` model can be dropped.

greatislander: Then user choices can be compared using numeric IDs instead of strings.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,93 @@
...
+            $userChoices = $data['question_'.$question->id];
+            if (count($userChoices) < $question->minimum_choices) {
+                continue;
+            } else {
+                $numberOfCorrectAnswers = 0;
+                foreach ($question->choices as $choice) {
+                    if (in_array($choice->value, $userChoices) && $choice->is_answer) {
+                        $numberOfCorrectAnswers++;
+                    }
+                }
```
greatislander: See if this comparison can be done using an array intersection of user choices and correct choices for a given question.

See: https://www.php.net/manual/en/function.array-intersect.php

greatislander:
@chosww I merged the latest changes in and now the tests are failing. Can you take a look?

jobara:
@chosww where do things stand on this PR? Is it ready for another round of review or are there still things that you are working on?

chosww:
hey @jobara, I am currently working on trainings dashboard page on a different branch off from this one, and I found some issues from this PR, and made fixes on the other branch. I am not sure if we should merge this in and I open another PR for the dashboard for your review or just close this PR and open a new PR that includes dashboard?

routes/courses.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use App\Http\Controllers\CourseController;
+
+Route::multilingual('/courses', [CourseController::class, 'index'])
+    ->name('courses.index');
```
jobara: Should the index also be guarded by the `auth` middleware? At the moment a guest user could access the listing of courses but would get a login request if they tried to view any.

app/Http/Controllers/CourseController.php
```diff
@@ -0,0 +1,33 @@
...
+            'courses' => Course::all(),
+        ]);
+    }
+
+    public function show(Course $course): View
+    {
+        $user = Auth::user();
+        $courseUser = $user->courses->find($course->id);
+        $finishedCourse = $courseUser?->getRelationValue('pivot')->finished_at;
+        $receivedCertificate = $courseUser?->getRelationValue('pivot')->received_certificate_at;
```
jobara: I'm not exactly sure what the `getRelationValue` method does here. Could this be written as the following?

```suggestion
        $finishedCourse = $courseUser?->pivot->finished_at;
        $receivedCertificate = $courseUser?->pivot->received_certificate_at;
```

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+use App\Http\Requests\StoreQuizResultRequest;
+use App\Models\Quiz;
+use Illuminate\Contracts\View\View;
+use Illuminate\Support\Facades\Auth;
+
+class QuizController extends Controller
+{
+    public function show(Quiz $quiz): View
+    {
+        $questions = $quiz->questions;
```
jobara: You don't actually use $questions for anything directly so you could skip this line and just return `$quiz->questions` below.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+            'questions' => $questions,
+        ]);
+    }
+
+    public function storeQuizResult(StoreQuizResultRequest $request, Quiz $quiz): View
+    {
+        $data = $request->validated();
+        $user = Auth::user();
+
+        $numberOfQuestions = 0;
```
jobara: You can get the count of the questions from the collection or relationship, so you don't need to manually increment a counter.

```suggestion
        $numberOfQuestions = $quiz->questions->count();
```

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        $quizScore = 0;
+
+        foreach ($quiz->questions as $question) {
+            $numberOfQuestions++;
+            if (count(array_intersect($question->getCorrectChoices(), $data['question_'.$question->id])) >= $question->minimum_choices) {
+                $quizScore++;
+            }
```
jobara: Could you explain what `minimum_choices` means. I would assume that to get the answer correct that `getCorrectChoices()` would have to have the same array values as was returned. However, it looks like they can answer some subset of correct choices and still get a pass. Also based on the logic here they could just select all of the options and they'd still pass the question too; which doesn't seem correct.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+        $quizResults = $quizScore / $numberOfQuestions >= $quiz->minimum_score;
+        $quizUser = $user->quizzes->where('id', $quiz->id)->first()->pivot ?? null;
+        if ($quizScore / $numberOfQuestions >= $quiz->minimum_score) {
```
jobara: If there are no questions I'm guessing you're going to get an error here because of the division by 0. I did take a look over the `StoreQuizResultRequest` class but couldn't quite tell if it was checking this.

jobara: Same for line 37 above.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        $quizScore = 0;
+
+        foreach ($quiz->questions as $question) {
+            $numberOfQuestions++;
+            if (count(array_intersect($question->getCorrectChoices(), $data['question_'.$question->id])) >= $question->minimum_choices) {
+                $quizScore++;
+            }
+        }
+        $quizResults = $quizScore / $numberOfQuestions >= $quiz->minimum_score;
+        $quizUser = $user->quizzes->where('id', $quiz->id)->first()->pivot ?? null;
```
jobara: The null coalescing is not needed here. The `??` means that if the left hand side is null return what is on the right hand side. What you are doing here in effect is checking if the `$user->quizzes->where('id', $quiz->id)->first()->pivot` is null then return null. So you could just remove it and you'll get the same value back.

```suggestion
        $quizUser = $user->quizzes->where('id', $quiz->id)->first()->pivot;
```

I'm not sure if I'm reading the logic here wrong or not, but it would also seem that you could simplify this because you are finding a specific quiz by id, you could use `find` and shouldn't need to call `first` as you are only getting one.

```suggestion
        $quizUser = $user->quizzes->find($quiz)->pivot;
```

It probably doesn't matter, but you might be able to go the opposite way. I wonder if this reads better because we are in the Quiz controller? Let me know what you think.

```suggestion
        $quizUser = $quiz->users()->find($user)->pivot;
```

app/Http/Controllers/QuizController.php
jobara: `$quizScore / $numberOfQuestions` this calculation is used multiple times and should be stored in a variable instead.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+                        'attempts' => 1,
+                        'score' => $quizScore / $numberOfQuestions,
+                    ]
+                );
+            }
+            $user->courses()->updateExistingPivot(
+                $quiz->course->id, [
+                    'received_certificate_at' => now(),
+                ]
+            );
```
jobara, line 76: It looks like this is the only difference between the if/else portions. I think you could remove the duplication to always run the other parts and just put this section on an if condition.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        foreach ($quiz->questions as $question) {
+            $numberOfQuestions++;
+            if (count(array_intersect($question->getCorrectChoices(), $data['question_'.$question->id])) >= $question->minimum_choices) {
+                $quizScore++;
+            }
+        }
+        $quizResults = $quizScore / $numberOfQuestions >= $quiz->minimum_score;
+        $quizUser = $user->quizzes->where('id', $quiz->id)->first()->pivot ?? null;
+        if ($quizScore / $numberOfQuestions >= $quiz->minimum_score) {
+            if ($quizUser) {
```
jobara: Under what circumstance would this be false? Is it that this is just true if the quiz is taken multiple times?

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+            $finishedCourse = true;
+            $course = $this->module->course;
+            foreach ($course->modules as $module) {
+                $moduleUser = $module->users->find($this->user->id)?->getRelationValue('pivot');
+                if (! $moduleUser) {
+                    $finishedCourse = false;
+                    break;
+                } elseif (! $moduleUser->finished_content_at) {
+                    $finishedCourse = false;
+                    break;
```
jobara: would be great if you could add a test that covers this case.

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+                }
```
jobara: I don't think you need an if/elseif here. You could either put both conditions into one if or have two separate if statements.

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+            }
+            if ($finishedCourse) {
+                $course->users()->updateExistingPivot(
+                    $this->user->id, ['finished_at' => now()]
+                );
+            }
+        }
```
jobara: I actually feel like this whole logic about deciding if a course is finished and updating the course status should actually be part of the course. You should probably move this to a method in the course that just gets called here, or perhaps you might be able to have some model observer in the Course that just triggers the thing automatically. I'm not sure if that's possible so some investigation would be needed.

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+
+    protected $listeners = ['onPlayerEnd', 'onPlayerStart'];
+
+    public function onPlayerEnd()
+    {
+        if (! $this->finishedContentAt) {
+            $this->user->modules()->updateExistingPivot(
+                $this->module->id, ['finished_content_at' => now()]
+            );
+            $this->finishedContentAt = 'finished_content';
```
jobara: I'm not sure what this is for and also why it's a string and not the time like is set in the pivot above.

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+                    $this->user->id, ['finished_at' => now()]
+                );
+            }
+        }
+    }
+
+    public function onPlayerStart()
+    {
+        if (! $this->startedContentAt) {
+            $course = $this->module->course()->first();
```
jobara: A module only has one course, so I think you could write this as:

```suggestion
            $course = $this->module->course;
```

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,78 @@
...
+            $this->user->modules()->attach(
+                $this->module->id, ['started_content_at' => now()]
+            );
+            $this->startedContentAt = 'started_content_at';
+        }
+    }
+
+    public function render()
+    {
+        return view('livewire.module-content')->layout('layouts.app-wide');
```
jobara: This is causing an error because `layouts.app-wide` doesn't exist anymore.

app/Http/Requests/StoreQuizResultRequest.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    // public function authorize()
+    // {
+    //     return false;
+    // }
```
jobara: This shouldn't be commented out. You can return `true` here if anyone is authorized, or more likely in this case you can return if there is an authenticated user.

app/Http/Requests/StoreQuizResultRequest.php
```diff
@@ -0,0 +1,35 @@
...
+        $questionRules = [];
+        $quiz = $this->route('quiz');
+        // @phpstan-ignore-next-line
+        foreach ($quiz->questions as $question) {
+            $questionRules['question_'.$question->id] = 'required|array';
+        }
+
+        return $questionRules;
+    }
+}
```
jobara: We might need to chat about this, I'm not sure I'm understanding this.

app/Mail/QuizResults.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Get the message envelope.
+     *
+     * @return \Illuminate\Mail\Mailables\Envelope
+     */
+    public function envelope()
+    {
+        return new Envelope(
+            subject: 'Quiz Results for '.$this->name,
```
jobara: Shouldn't this be localized?

app/Models/Question.php
```diff
@@ -36,4 +43,24 @@ public function choices(): HasMany
     {
         return $this->hasMany(Choice::class);
     }
+
+    public function getChoices()
+    {
+        return Options::forModels(
+            Choice::query()->whereIn('id', $this->choices->pluck('id')->toArray()),
+            label: fn (Choice $choice) => $choice->getTranslation('label', 'en')
```
jobara: I think this is going to force the translation to be in English. I think you don't need the second argument.

app/Models/Question.php
```diff
@@ -36,4 +43,24 @@ public function choices(): HasMany
     {
         return $this->hasMany(Choice::class);
     }
+
+    public function getChoices()
+    {
+        return Options::forModels(
+            Choice::query()->whereIn('id', $this->choices->pluck('id')->toArray()),
```
jobara: I think you can use [modelKeys](https://laravel.com/docs/9.x/eloquent-collections#method-modelKeys) to write the following as:
```suggestion
            Choice::query()->whereIn('id', $this->choices->modelKeys()),
```

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,72 @@
...
+        </div>
+        @if ($finishedCourse && !$receivedCertificate)
+            <form class="stack" action="{{ localized_route('quizzes.show', $course->quiz) }}" novalidate>
+                @csrf
+                <button>{{ __('Take Quiz') }}</button>
+            </form>
+        @elseif(!$finishedCourse && !$receivedCertificate)
+            <button type="button" @ariaDisabled>{{ __('Take Quiz') }}</button>
+        @elseif($receivedCertificate)
+        @endif
```
jobara, line 81: Is something missing here?

routes/modules.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+use App\Http\Livewire\ModuleContent;
+
+Route::multilingual('/modules/{module}', [ModuleContent::class, '__invoke'])
```
jobara: should probably be under courses. that is `/courses/{course}/{module}` or `/courses/{course}/modules/{module}`

routes/quizzes.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+use App\Http\Controllers\QuizController;
+
+Route::multilingual('/quizzes/{quiz}', [QuizController::class, 'show'])
```
jobara: This should probably be under `courses/{course}/quiz` and possibly one similar to this but for modules. It seems there can only be one quiz so you probably don't need to pass in the ID, but if you need to you could do that too I guess.

routes/quizzes.php
```diff
@@ -0,0 +1,12 @@
...
+    ->middleware('auth')
+    ->name('quizzes.show');
+
+Route::multilingual('/quizzes/{quiz}/result', [QuizController::class, 'storeQuizResult'])
+    ->method('post')
+    ->middleware('auth')
+    ->name('quizzes.show-result');
```
jobara: I think the name here must be wrong as this appears to store the result, but the name is `show`.

jobara: Another thing is you should follow the pattern in other route files to use groups.

jobara (changes requested):

jobara:
@chosww you can use the same videos that are used in the [InterpreationFactory](https://github.com/accessibility-exchange/platform/blob/9a84f28a8ee3c54eabc23311445799e8db9f4039/database/factories/InterpretationFactory.php#L24-L25) to add videos to your training factories.

app/Http/Livewire/ModuleContent.php
jobara: Another thought is to not have this in the pivot at all, unless the `finished_at` time is actually important but even then it could be defined from the latest module finished at time. Also there is some ambiguity here because the course is marked as finished when the modules are complete but not when the quiz is complete. So maybe you just have a function that you fun to see if all the courses modules are complete for the user. This method could be used for the policy checks I mention elsewhere too. Another thing with the existing code is that if a module is added after the user has finished the existing modules, it will still be marked as complete.

jobara: It's probably also worth investigating to see if you can find only the modules that don't have a 'finished_content_at' in the module_user pivot. That way you can just see if the count is 0 to know if course is complete.

routes/courses.php
chosww: We don't need an index page for the course.

app/Http/Controllers/CourseController.php
chosww: I used getRelationValue to get around the composer analysis error saying that pivot is not defined for relationships. This method was also suggested by @greatislander.

app/Http/Controllers/QuizController.php
chosww: My concern was that calling quiz->questions would call another query. @jobara and I discussed about this, and we could potentially store $quiz->questions to a variable, then get how many questions are there, or use [withCount](https://laravel.com/docs/9.x/eloquent-relationships#counting-related-models) to get the counts.

app/Http/Controllers/QuizController.php
chosww: @michelled could we talk about this when you and @jobara are available?

app/Http/Livewire/ModuleContent.php
chosww: I agree that we don't need finished_at attribute for the course, we could have a method similar to isPublishable, where it iterate through modules in courses and check their finished status to decide if a course is finished or not.

resources/views/courses/show.blade.php
chosww: In another work I started off from this branch populates texts in this block, so let's keep this block on this PR.

routes/modules.php
chosww: decided to use the first route.

chosww: Hey @jobara, I couldn't figure out how I can do route model binding for {course} from modules route. Could you help me on this please?

jobara: @chosww could you explain a bit more of what the issue you are running into is? What part isn't working. I'm not sure if it's an issue with the livewire component for example or maybe the way you are calling the route?

For example when you're [accessing the route](https://github.com/accessibility-exchange/platform/blob/1325a300f43de162e2e645f8b5bbb6d06435da81/resources/views/courses/show.blade.php#L46) you'll need to pass in the course instance. 

Something like:

```php
<a href="{{ localized_route('modules.module-content', ['course' => $module->course, 'module' => $module]) }}">
    {{ $module->title }}
</a>
```

You may also want to look at [scoped bindings](https://laravel.com/docs/9.x/routing#implicit-model-binding-scoping).

jobara: If you're still running into issues, it may be something that needs to be updated in the livewire `mount` to take in the `course` or you may need to look at this livewire issue, https://github.com/livewire/livewire/issues/2558, where they talk about explicitly setting the model binding.

chosww: Thanks Justin, I was able to resolve the issue I was having with your first suggestion.

chosww:
Hey @jobara I updated quiz validation to match the work flow in the design, so that when users have not passed the quiz, it will show incorrect questions users have answered. However, I am having some trouble keeping the selected choices persist with hearth checkboxes component. Could we chat about the issue once you come back from your vacation?

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    public function show(Course $course): View
+    {
+        $quiz = $course->quiz;
+
+        return view('quizzes.show', [
+            'course' => $course,
+            'quiz' => $quiz,
+            'title' => $quiz->title,
+            'questions' => $quiz->questions,
```
jobara: you are passing back the `quiz` and then also parts of the `quiz` for the `title` and `questions`. It doesn't look like you use the whole `quiz` instance in the template, so you might be able to just omit this. You may also need to do [lazy eager loading](https://laravel.com/docs/9.x/eloquent-relationships#lazy-eager-loading) on the questions.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        ]);
+    }
+
+    public function storeQuizResult(StoreQuizResultRequest $request, Course $course): View | RedirectResponse
+    {
+        $data = $request->validated();
+        $user = Auth::user();
+        $correctQuestions = 0;
+        $quiz = $course->quiz;
+        $quizWithCounts = $quiz->loadCount('questions');
```
jobara: I could be wrong but I don't think you need to save this to a new variable. I think `loadCount` will update `$quiz` to include the count.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+            $previousAnswers[$question->id] = $data['questions'][$question->id];
+            if ($question->getCorrectChoices() == $data['questions'][$question->id]) {
+                $correctQuestions++;
+            } else {
+                $validator->errors()->add('questions.'.$question->id, __('Wrong answer'));
+                $wrongAnswers[] = $question->id;
+            }
+        }
+
+        //@phpstan-ignore-next-line
```
jobara: What does the next line need to be ignored?

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+        if ($quizWithCounts->questions_count > 0) {
```
jobara: It would seem that this condition should be checked at the top and return immediately if the question_count is 0. For example the foreach wouldn't have any questions to iterate over either. I'm not sure if this should throw an error because in reality this controller function shouldn't be run at all if there are no questions.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,82 @@
...
+                    $quiz->course->id, [
+                        'received_certificate_at' => now(),
+                    ]
+                );
+            } else {
+                return redirect(localized_route('quizzes.show', $course))->with(['previousAnswers' => $previousAnswers, 'wrongAnswers' => $wrongAnswers])->withErrors($validator);
+            }
+        }
+
+        return view('quizzes.store-result', ['quiz' => $quiz, 'results' => $isPass]);
```
jobara: It seems like this route has the wrong name. I would think it would be something like `'quizzes.show-result'`.

app/Http/Controllers/QuizController.php
jobara: At the moment what will happen is that the test will go to the results page with a failure result. In other cases of failure that page is redirected back to the quiz with info on incorrect answers.

app/Http/Livewire/ModuleContent.php
```diff
@@ -0,0 +1,61 @@
...
+            );
+            $this->isFinished = true;
+        }
+    }
+
+    public function onPlayerStart()
+    {
+        if (! $this->isStarted) {
+            $course = $this->module->course;
+            if (! ($this->user->courses->find($course->id)?->getRelationValue('pivot'))) {
```
jobara: Do you need to check the presence of the pivot to determine if it's attached? Would the find be sufficient?

app/Http/Requests/StoreQuizResultRequest.php
```diff
@@ -0,0 +1,43 @@
...
+     * Get the validation rules that apply to the request.
+     *
+     * @return array<string, mixed>
+     */
+    public function rules()
+    {
+        $rules = [];
+        foreach ($this->course->quiz->questions as $question) {
+            $rules['questions.'.$question->id] = 'required';
+        }
```
jobara, line 29: Sorry I forgot to ask about this when we were talking the other day. You mentioned you had issues using `questions.*`. Could you explain what didn't work with that?

app/Http/Requests/StoreQuizResultRequest.php
```diff
@@ -0,0 +1,43 @@
...
+
+        return $rules;
+    }
+
+    public function messages()
+    {
+        $messages = [];
+        foreach ($this->course->quiz->questions as $question) {
+            $messages['questions.'.$question->id.'.required'] = __('You must answer this question');
+        }
```
jobara: I would have imagined you could have used `questions.*` here as well.

app/Mail/QuizResults.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Get the message envelope.
+     *
+     * @return \Illuminate\Mail\Mailables\Envelope
+     */
+    public function envelope()
+    {
+        return new Envelope(
+            subject: __('Quiz Results for ').$this->name,
```
jobara: Probably better to write this as 

```suggestion
    {
        return new Envelope(
            subject: __('Quiz Results for :name ', ['name' => $this-name]),
```

In doing so you allow for more flexibility in the localization. That is, the localized version may want to have the name appear in a different part of the string or perhaps with some punctuations or additional content after the name.

app/Models/Course.php
```diff
@@ -51,4 +53,20 @@ public function modules(): HasMany
...
+    {
+        $isFinished = true;
+        foreach ($this->modules as $module) {
+            $moduleUser = $module->users->find($user->id)?->getRelationValue('pivot');
+            if ($moduleUser?->finished_content_at) {
+                continue;
+            } else {
+                $isFinished = false;
+                break;
+            }
```
jobara: You can do this with just an if statement if you use a negated condition `if (! $moduleUser?->finished_content_at)`. Although be aware that if $moduleUser is null you will consider the course finished. Is that what you want?

Also you may want look at [Filtering Queries Via Intermediate Table Columns](https://laravel.com/docs/9.x/eloquent-relationships#filtering-queries-via-intermediate-table-columns) to see if you could build this differently so you don't have to loop through all of the modules but could just query to see if any of them have `finished_content_at` as `null`.

app/Models/Question.php
```diff
@@ -36,4 +37,24 @@ public function choices(): HasMany
...
+    public function getCorrectChoices()
+    {
+        $correctChoices = [];
+        foreach ($this->choices as $choice) {
+            if ($choice->is_answer) {
+                $correctChoices[] = $choice->id;
+            }
+        }
+
+        return $correctChoices;
```
jobara: I'm having some second thoughts about the structure of quiz, question and choice all being separate models. Could you give me a quick summary of the reasoning for the structure.

database/migrations/2022_11_21_194859_add_minimum_choices_to_questions_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
jobara: As we are no longer using `minimum_choices` this migration can be removed.

database/migrations/2022_12_20_162007_remove_minimum_choices_from_questions_table.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
jobara: As we are no longer using `minimum_choices` this migration can be removed.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,73 @@
...
+        <div class="flex justify-between">
+            <div>
+                {{ __('Author') }}
+            </div>
+            <div>
+                {{ __('Resource type') . ': ' . __('Training') }}
+            </div>
+            <div>
+                {{ __('Published on') }}
+            </div>
```
jobara: I think the values for the Author and Published dates are missing.

resources/views/emails/quiz/results.blade.php
```diff
@@ -0,0 +1,9 @@
+<x-mail::message>
+    # {{ __('Quiz results') }}
+
+    {{ __('Congraturations') . ', ' . $name . '!' }}
+    {{ __('You have successfully completed course') . ' ' . $course . ' ' . __('and passed the quiz') . '.' }}
```
jobara: These should all use the string interpolation. Also there is a typo Congraturations => Congratulations.

```suggestion
    {{ __('Congratulations, :name!', ['name' => $name]) }}
    {{ __('You have successfully completed course :course and passed the quiz.', ['course' => $course] }}
```

resources/views/livewire/module-content.blade.php
```diff
@@ -0,0 +1,28 @@
+<x-slot name="title">
+    {{ __('Module') . '-' . $module->title }}
```
jobara:
```suggestion
    {{ __('Module - :title', ['title' => $module->title]) }}
```

Please review all string concatenations throughout.

resources/views/quizzes/show.blade.php
```diff
@@ -0,0 +1,33 @@
...
+                        @if (in_array($question->id, session('wrongAnswers')))
+                            <x-banner type="error">
+                                {{ __('Please try again.') }}
+                            </x-banner>
+                        @else
+                            <x-banner type="success">
+                                {{ __('Correct answer!') }}
+                            </x-banner>
+                        @endif
+                    @endif
```
jobara: This doesn't appear to be valid markup. The `<legend>` element needs to be the first child of a `<fieldset>`. I suppose you could move the banners above the `<fieldset>` or below `<legend>` depending on where you want them to show up.

see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend

resources/views/quizzes/show.blade.php
```diff
@@ -0,0 +1,33 @@
...
+                    <legend>{{ $question->question . '?' }}</legend>
+                    <x-hearth-checkboxes name="questions[{{ $question->id }}]" :options="$question->getChoices()" :checked="session('previousAnswers') && array_key_exists($question->id, session('previousAnswers'))
+                        ? session('previousAnswers')[$question->id]
+                        : []" />
```
jobara: what happens if the validation fails for a reason other than the answer being incorrect. I'm guessing you did want the old value and this as the fallback. Something like the following:

```suggestion
                    <x-hearth-checkboxes name="questions[{{ $question->id }}]" :options="$question->getChoices()" :checked="old(questions[{{ $question->id }}], session('previousAnswers') && array_key_exists($question->id, session('previousAnswers'))
                        ? session('previousAnswers')[$question->id]
                        : [])" />
```

tests/Feature/QuizTest.php
```diff
@@ -42,3 +36,178 @@
...
+    $secondQuestionWrongChoice = Choice::factory()->for($secondQuestion)->create();
+
+    $response = $this->actingAs($user)->get(localized_route('quizzes.show', $course));
+    $response->assertOk();
+    $response->assertSee($firstQuestion->title);
+    $response->assertSee($firstQuestionCorrectChoice->label);
+    $response->assertSee($firstQuestionWrongChoice->label);
+    $response->assertSee($secondQuestion->title);
+    $response->assertSee($secondQuestionCorrectChoice->label);
+    $response->assertSee($secondQuestionWrongChoice->label);
```
jobara: You may want to use [`assertSeeInOrder`](https://laravel.com/docs/9.x/http-tests#assert-see-in-order) instead.

app/Http/Controllers/QuizController.php
jobara: I'm was playing around with the training. It occurs to me that because we allow the quiz to pass with less than 100% correct, users won't know that they've made mistakes. We may always want to redirect back to the test to show the results of their answers if they pass but in those cases show what the correct answers should be. I think we may need to talk to @michelled and @cherylhjli about this.

jobara (changes requested):

app/Http/Controllers/QuizController.php
chosww: Composer analyze throws an error "Access to an undefined property App\Models\Quiz::$questions_count", so I had to place this ignore next line.

app/Http/Livewire/ModuleContent.php
chosww: Yeah, it seems like there's no need for checking the presence of the pivot.

app/Http/Requests/StoreQuizResultRequest.php
chosww: So, questions.* would work if the key within the questions doesn't matter, but we want to make sure all specific questions have answered. For an example questions could have key 1 and 3, (the keys are the ids of the questions), and answering just one of them would be qualified for questions.*, as it will store answer to a question to the questions array. But what we want is that both questions[1] and questions[3] have answered, so that's why I wrote a logic to check all the keys in questions have answered.

app/Models/Question.php
chosww: It's been a while so I don't quite remember clearly, but while working on this task, Ned and I realized that it would be more convenient for us to have more granular model designs. Is there better design you want to explore?

app/Http/Controllers/QuizController.php
jobara: Seems there are two options here for now. 

1. leave as you have it but add an additional comment above to indicate why that line is being ignored. You should also reference this issue: https://github.com/nunomaduro/larastan/issues/466
2. You can add `/** @property int $questions_count */` above line 12 of App/Models/Quiz which will tell larastan that  it is a valid property.

app/Http/Requests/StoreQuizResultRequest.php
jobara: I wonder if the following would work.

```php
[
    'questions' => [
        'required',
        'array',
        'size:'.$this->course->quiz->questions->count()
    ],
]
```

app/Models/Question.php
jobara: Well that's the question I was hoping to figure out. I'm trying to understand the separation of concerns here and wondering if there were future plans or access requirements for the structure. For example, would we want to reuse questions or choices and make them many-to-many relationships. One thing I'm sort of finding a bit strange is that the question doesn't know the correct answer, but the choice knows if it is correct.

For example if we had a questions like "Who should you contact if you have further concerns?". Suddenly the answer to the question depends on the context (i.e the quiz). The correct choice depends on the question which depends on the quiz.

Now let's say that we will never re-use questions and choices, so they are always a one-to-many relationship. Can we just make them an array of options on the question model? Is it for localization that they are separate models? Should the question really be the one that knows the correct answer?

jobara: To summarize from the discussion we had yesterday at the IDRC planning meeting:

- General suggestion to make the system more flexible, especially because we don't really know the full requirements
- Allow for quizzes at the course and/or module level
  - This will mean that we need to determine when the modules are complete by checking if the the videos have finished and if the available module quizzes have been passed.
  - The course is complete if all modules are complete and if there is a course quiz that it has passed
- The questions should be available to be added to any quiz, many-to-many relationship 
- The choices will be part of a question as opposed to being in a separate table. There will be an array of choices, and an array of correct answers which is an array of the indexes of the choices that are correct. Both of these arrays are required.
  - Need to make sure that we can localize the choices.

app/Http/Controllers/QuizController.php
chosww: Thanks Justin, I took the second approach you suggested.

app/Http/Controllers/QuizController.php
chosww: I just removed this check, because we are already checking if there's at least 1 question in the quiz before interacting with this controller method.

app/Http/Controllers/QuizController.php
chosww: I set the default pass threshold to be 100% in Quiz model, so unless users get perfect score on the quiz, they won't get the certificate and see the pass page.

app/Http/Requests/StoreQuizResultRequest.php
chosww: I wasn't able to get correct error messages this way, so I didn't make this changes.

resources/views/courses/show.blade.php
chosww: Ned told me before there may be a case where Author is unclear, when an org uploads a course created by different party, we can talk about how we will approach in that case, but I just commented the author part for now.

app/Models/Course.php
chosww: I took the first approach with additional logic to return false if there's no training modules. I couldn't figure out using the intermediate table queries, and I think it's because this is Course model, and $this would return a course instance.

resources/views/courses/show.blade.php
jobara: Maybe something to talk about with @cherylhjli. I believe the author could be pulled from either the Author field or a linked organization. In which case you could use the author property or fallback to the name of a linked organization. Would that satisfy the concern?

app/Http/Controllers/QuizController.php
jobara: Hmm.. I'm not sure that will actually solve the issue, in the case where the course creator lowers the threshold the user can still pass without realizing the mistakes or knowing the correct answer. Perhaps we should talk to @cherylhjli and @michelled about this one again. If it's too complicated to make the change to always go back to the quiz to show the correct answers, maybe we shouldn't allow the threshold to be anything other than 100%. However, this will impact the user a bit in regards to being more likely to hit the number of attempts limit.

chosww:
We talked about possible Module level quiz: #1515.

jobara:
@chosww could you please resolve the merge conflict for this PR?

app/Models/Question.php
```diff
@@ -16,24 +15,24 @@ class Question extends Model
     protected $fillable = [
         'order',
```
jobara: Now that a question could be attached to multiple quizzes, specifying the order in the question wouldn't work, as it may be in different locations in each quiz. A quiz should maintain the order of its questions.

app/Http/Controllers/QuizController.php
```diff
@@ -0,0 +1,72 @@
...
+            $isPass = true;
+            $user->courses()->updateExistingPivot(
+                $quiz->course->id, [
+                    'received_certificate_at' => now(),
+                ]
+            );
+        }
+
+        return redirect(localized_route('quizzes.show', $course))->with(['previousAnswers' => $previousAnswers, 'wrongAnswers' => $wrongAnswers, 'isPass' => $isPass, 'score' => $quizScore])->withErrors($validator);
+    }
```
jobara, line 81: Because we redirect back to the 'quizzes.show' route, if the user refreshes the page, they will be put back into the mode of taking the quiz. I suppose this would also be the case if they access the URL for the quiz directly. However, you'd probably want to redirect to the result instead. 

Of course we aren't storing the answers and errors so those would be lost. Might want to talk to @cherylhjli about what we want here. For example we might just show the score and buttons for going back and e-mailing the certificate.

app/Http/Requests/StoreQuizResultRequest.php
```diff
@@ -0,0 +1,43 @@
...
+
+        return $messages;
```
jobara: I believe you are using the function to define the rules because you want to explicitly validated each question to be present and show the error message with the corresponding UI element. However; in the case of the messages I think you can write this using the standard notation.

```suggestion
        return [
            'questions.*' =>  __('You must answer this question');
        };
```

app/Mail/QuizResults.php
jobara: You should review the string concatenations through out. Anything that is sent for translation should interpolate as the suggestion above.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,73 @@
...
+            {{ $course->title }}
+        </h1>
+    </x-slot>
+    <div class="stack ml-2 mr-2">
+        <a href="{{ localized_route('resource-collections.index') }}">{{ __('Back') }}</a>
+        <x-slot name="title">{{ $course->title }}</x-slot>
+        <div class="flex justify-between">
+            {{-- <div>
+                {{ __('Author') }}
+            </div> --}}
```
jobara: We still need to show the author. The course model should be like Resources where the author is either pulled from an 'author' property or the related organization.

resources/views/quizzes/show.blade.php
```diff
@@ -0,0 +1,100 @@
...
+    <x-slot name="header">
+    </x-slot>
+    <div class="stack ml-2 mr-2">
+        <x-slot name="title">{{ $title }}</x-slot>
+    </div>
+    <div class="stack ml-2 mr-2">
+        @if (session('isPass') !== null)
+            <h1>{{ __('Quiz results') }}</h1>
+            @if (session('isPass'))
+                <h3>{{ __('Congratulations! You have passed the quiz with :score%.', ['score' => session('score') * 100]) }}
```
jobara: The score should be rounded. For example to the second decimal place so you may have something like "66.67%" instead of something like "66.666666666666666667%". We may want to check with @cherylhjli to see exactly how precise the rounding should be. 

See: [round()](https://www.php.net/manual/en/function.round.php)

jobara: The percentage score should likely be calculated in the controller instead of in the template.

jobara (changes requested):

jobara:
The modules are missing the `.card` styling. 
https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=1368%3A19472&t=OtcCyjOb1zLI4bR5-0

Screenshot of module cards without the card styling

jobara:
Missing the button to resend the certificate from the course landing page.
https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=1368%3A21765&t=OtcCyjOb1zLI4bR5-0

jobara:
Could you please add a quick login button for a training consultant. Similar to what is available for the Individual user and etc.

Screenshot of quick login options

app/Models/Quiz.php
```diff
@@ -34,21 +36,32 @@ class Quiz extends Model
...
+            $orderedQuestions = collect([]);
+            foreach ($this->order as $id) {
+                $orderedQuestions->push(Question::find($id));
+            }
+
+            return $orderedQuestions->merge($this->questions)->unique('id')->values();
+        }
+
+        return $this->questions;
     }
```
jobara, line 66: I wonder if you could use [`sort`](https://laravel.com/docs/9.x/collections#method-sort) or [`sortBy`](https://laravel.com/docs/9.x/collections#method-sortby) with a function passed in that returns the index in the order array? Whichever approach is more performant is fine.

jobara (changes requested):

app/Models/Quiz.php
chosww: Hey @jobara when I think about your question again, the lines 56 to 60 are to generate a Question collection with given question ids from $this->order; the logic is not to sort the collection, but generated collection is sorted. The line 62 is to remove the duplication on the collection.

chosww: Is it possible to use sort or sortBy to generate a collection with given array of ids?

jobara: I think you'd end up doing something like:

```php
$this->questions->sort(function ($a, $b) {
    if ($a->id === $b->id) {
        return 0;
    }

    $aIndex = array_search($a->id, $this->order);
    $bIndex = array_search($b->id, $this->order);

    if ($aIndex === $bIndex) {
        return 0;
    }

    if ($aIndex === false) {
        return 1;
    }

    if ($bIndex === false) {
        return -1;
    }

    return $aIndex < $bIndex ? -1 : 1; })
});
```

That looks a bit complicated, but maybe you could optimize it a bit. Not sure.

chosww: Thanks @jobara, the sort function seems to work for the test case I wrote for the function, but not on the UI; I am not sure what causes the difference what's appearing on the UI, but the questions are not ordered according to the order property, but they are ordered correctly by id.

resources/views/courses/show.blade.php
```diff
@@ -0,0 +1,81 @@
...
+        <x-slot name="title">{{ $course->title }}</x-slot>
+        <div class="flex justify-between">
+            <div>
+                {{ __('Author: :author', ['author' => $course->author ?? $course->organization?->name]) }}
+            </div>
+            <div>
+                {{ __('Resource type: Training') }}
+            </div>
+            <div>
+                {{ __('Published on: :created_at', ['created_at' => $course->created_at->isoFormat('MM, DD, YYYY')]) }}
```
jobara: The date is coming up as 02, 23, 2023 instead of February 23, 2023. I think you mean to use the following:

```suggestion
                {{ __('Published on: :created_at', ['created_at' => $course->created_at->isoFormat('LL')]) }}
```

jobara:
The "In progress" badge doesn't appear for modules. 

I opened a module, started a video and paused before it finished. I went back to the course page and there wasn't a budge on the module.

see: https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=6743%3A113461&t=OtcCyjOb1zLI4bR5-0

app/Models/Quiz.php
jobara: @chosww could be because the sorted array keeps the original keys. You would need to call the `values` method to get the keys to be reindexed.

```php
$sorted = $this->questions->sort(...);
$sorted->values()->all();
```

jobara (changes requested):

jobara (approved):

PR #1473, opened 2023-01-13
Feat: tag asl and lsq contents (resolves #1443)

### LINKED ISSUE #1443: Add ASL and LSQ for two flows

chosww:
Resolves #1443 

### Prerequisites


resources/views/individuals/partials/progress.blade.php
```diff
@@ -27,6 +27,7 @@
                 @endcan
                 <button class="secondary" name="publish" value="1"
                     @cannot('publish', $individual) @ariaDisabled aria-describedby="cannot-publish-explanation" @endcannot>{{ __('Publish page') }}</button>
+                <x-interpretation class="interpretation--start" name="{{ __('Publish page', [], 'en') }}" />
```
jobara: I'm pretty sure that `'Publish page'` will be used in future pages. Can you please add a namespace to this one? Please also review the other interpretations to see if any of those will be re-used on other pages as well. Might as well add namespaces to all the ones necessary now if we can.

resources/views/components/translatable-input.blade.php
```diff
@@ -7,6 +7,7 @@
                 @else
                     <x-hearth-label :for="$name . '_' . $language" :value="$label . ' (' . get_language_exonym($language) . ')'" />
                 @endif
+                <x-interpretation class="interpretation--start" name="{{ __($label, [], 'en') }}" />
```
jobara: Unfortunately this won't work as is because the `$label` has already passed through localization and will be the localized text. You'll unfortunately need to pass in the unlocalized text. I'm guessing you'll need to update the translatable components to take in two new optional parameters the interpretation name and namepace. If the interpretation name is provided then the `<x-interpretation>` component is conditionally added. 

The same goes for the translatable textarea.

jobara: Another option would be to require the input to the translatable inputs not be localized and to handle that internally like the interpretation component does. However, I think that may be more confusing for usage and would require them all to be updated now. Let me know what you think.

jobara: And another alternative would be to change the x-interpretation component so that it had separate instances for English/ASL and French/LSQ. However, this would make it more difficult to match up videos in the admin interface as you'd have to separately find them in English and French.

resources/views/individuals/partials/progress.blade.php
jobara: Also need to add the version for when the page isn't in a publishable state.

resources/views/individuals/edit/about-you.blade.php
```diff
@@ -12,11 +12,13 @@
                 {{ __('Step :current of :total', ['current' => request()->get('step') ?? 1, 'total' => $individual->isConnector() ? 5 : 4]) }}<br />
                 {{ __('About you') }}
             </h2>
+            <x-interpretation class="interpretation--start" name="{{ __('About you', [], 'en') }}" />
 
             <hr class="divider--thick">
 
             <div class="field @error('name') field--error @enderror">
                 <x-hearth-label for="name"><x-required>{{ __('Name') }}</x-required></x-hearth-label>
+                <x-interpretation class="interpretation--start" name="{{ __('Name', [], 'en') }}" />
```
jobara: The sign language videos include if a field is optional or required. If the `<x-required>` or `<x-optional>` inputs are used for the field we should indicate this in the name of the interpretation. Please review throughout.

```suggestion
                <x-interpretation class="interpretation--start" name="{{ __('Name (required)', [], 'en') }}" />
```

jobara: We might also want to add something to the name to indicated that it is for a form field as opposed to some other part of the text. To avoid name collision confusion when we have more translations. Maybe something like "Form field: Name (required)". I'm not sure that's the best approach though, so welcome to other ideas/suggestions.

resources/views/individuals/edit/about-you.blade.php
```diff
@@ -99,6 +111,9 @@
             <div class="field @error('website_link') field-error @enderror">
                 <x-hearth-label class="h4"
                     for="website_link"><x-optional>{{ __('Website link') }}</x-optional></x-hearth-label>
+                @if (locale() === 'lsq')
+                    <x-interpretation class="interpretation--start" name="{{ __('Website link', [], 'en') }}" />
+                @endif
```
jobara: This shouldn't only be for LSQ

resources/views/individuals/edit/communication-and-consultation-preferences.blade.php
```diff
@@ -97,16 +117,32 @@
...
                 <x-hearth-error for="meeting_types" />
             </fieldset>
 
             <hr class="divider--thick">
 
             <p class="flex flex-wrap gap-7">
-                <button class="secondary" name="save_and_previous"
-                    value="1">{{ __('Save and previous') }}</button>
-                <button name="save" value="1">{{ __('Save') }}</button>
+                @if (locale() === 'asl' || locale() === 'lsq')
```
jobara, line 136: This should probably also check if there are videos available. That is only when the `<x-interpretation>` component is rendered. However, practically speaking it should be okay in this case because we do have the videos for these so we should be able to expect that they are added in.

resources/views/layouts/footer.blade.php
```diff
@@ -26,6 +26,8 @@
             <div class="switcher grow:2">
                 <div class="stack" id="contact" tabindex="-1">
                     <h2>{{ __('Contact') }}</h2>
+                    <x-interpretation class="interpretation--start" name="{{ __('Contact', [], 'en') }}"
+                        namespace="contact" />
```
jobara: May need to modify this name so that it is distinguishable from other contact information. Maybe something like "Site Contact" or "Footer: Contact", etc.

resources/views/livewire/experiences.blade.php
```diff
@@ -4,37 +4,45 @@
...
 
                 <button class="secondary" type="button"
                     wire:click="removeExperience({{ $i }})">{{ __('Remove experience') }}</button>
+                <x-interpretation class="interpretation--start" name="{{ __('Remove experience', [], 'en') }}" />
             </li>
         @endforeach
     </ul>
     @if ($this->canAddMoreExperiences())
         <button class="secondary" type="button" wire:click="addExperience">{{ __('Add experience') }}</button>
+        <x-interpretation class="interpretation--start" name="{{ __('Add experience', [], 'en') }}" />
```
jobara: This is getting added twice. Once is from the individuals.edit and another from livewire.message. Probably need to add a namespace to this one.

jobara (changes requested):
There are few missing entries that I couldn't mark in the code review. 

- In welcome.blade.php, the "The Accessibility Exchange" has two possible interpretations; one with and one without the "Sign up"; which is only shown for guests.
- In progress.blade.php missing the interpretation for the "Page Sections". There are two versions of this depending if the individual has a community connector role or not.
- In about-you.blade.php missing the "City or town (optional)" interpretation.
- In for-individuals.blade.php the "Join our accessibility community" interpretation should be updated to be a new one. As this is a longer version and is different from the one on the welcome page.

resources/views/individuals/edit/about-you.blade.php
chosww: Thanks @jobara, I was thinking, if an interpretation is for a form field, wouldn't it have either "required" or "optional" in the name? If that's the case, I think we can have two different instances, for example "Name (required)" and "Name".

resources/views/components/translatable-input.blade.php
```diff
@@ -7,7 +7,10 @@
                 @else
                     <x-hearth-label :for="$name . '_' . $language" :value="$label . ' (' . get_language_exonym($language) . ')'" />
                 @endif
-                <x-interpretation class="interpretation--start" name="{{ __($label, [], 'en') }}" />
+                @if ($interpretationName)
+                    <x-interpretation class="interpretation--start" name="{{ __($interpretationName, [], 'en') }}"
+                        namespace={{ $interpretationNameSpace }} />
+                @endif
```
jobara: Missing the "" marks around the `namespace` value.

app/View/Components/TranslatableInput.php
```diff
@@ -56,12 +56,26 @@ class TranslatableInput extends Component
...
+     */
+    public $interpretationName;
+
+    /**
+     * The namespace for asl or lsq content.
+     *
+     * @var bool
+     */
+    public $interpretationNameSpace;
+
```
jobara: Should make these nullable and they are strings not boolean values. `@var string` and `public ?string $interpretationName` and `public ?string $interpretationNameSpace`.

app/View/Components/TranslatableInput.php
```diff
@@ -56,12 +56,26 @@ class TranslatableInput extends Component
...
     /**
      * Create a new component instance.
      *
      * @return void
      */
-    public function __construct($name, $label, $hint = null, $model = null, $shortLabel = '', $required = false)
+    public function __construct($name, $label, $hint = null, $model = null, $shortLabel = '', $required = false, $interpretationName = '', $interpretationNameSpace = '')
```
jobara: Should make the $interpretationName and $interpretationNameSpace default to `null`.

resources/views/components/translatable-input.blade.php
jobara: Defaulting `$interpretationName` and `$interpretationNameSpace` to `null` can used `@isset` instead of `@if`. You can also used this internally to set the `namespace` value only when `$interpretationNameSpace` or allow it to pass down `null` but would need to make sure it doesn't pass the string "null".

resources/views/about/for-individuals.blade.php
```diff
@@ -79,7 +79,7 @@
             <x-section class="full accent" aria-labelledby="join">
                 <div class="center center:wide stack stack:xl">
                     <h2 class="text-center" id="join">{{ __('Join our accessibility community') }}</h2>
-                    <x-interpretation name="{{ __('Join our accessibility community', [], 'en') }}" namespace="join" />
+                    <x-interpretation name="{{ __('Join our accessibility community (individuals)', [], 'en') }}" />
```
jobara: Might want to keep the name the as "Join our accessibility community" and just remove the namespace or change the namespace. Something that just occurred to me is that the localization strings won't be updated with unique names in this field. I think we should update the admin panel for the interpretations to show the namespace in the table view. We should probably have a chat about this when you're looking at things to work out the best approach.

resources/views/individuals/edit/about-you.blade.php
jobara: I left a comment about this in my upcoming review in relation to another change, but I just remembered that the interpretation name doesn't get added to the localization files because it is dynamically used. So these unique names won't be localized. It isn't a huge issue at the moment because the admin panels aren't localized, but once they are these will break. I'm thinking we should make this info to the namespace instead, but we talk it through.

resources/views/welcome.blade.php
```diff
@@ -3,19 +3,18 @@
...
         <div class="center center:wide stack stack:lg">
             <h1 itemprop="name">{{ __('The Accessibility Exchange') }}</h1>
-            <x-interpretation name="{{ __('The Accessibility Exchange', [], 'en') }}" />
             <p class="h4">
                 {{ __('Connecting the disability and Deaf communities and their supporters with ') }}<br />{{ __('organizations and businesses to work on accessibility projects together.') }}
             </p>
             @guest
                 <p><a class="cta" href="{{ localized_route('register') }}"> {{ __('Sign up') }}</a></p>
             @endguest
+            <x-interpretation name="{{ __('The Accessibility Exchange', [], 'en') }}" />
```
jobara: There are two versions of this interpretation, one that includes "Sign up" for guests and one that doesn't for authorized users.

resources/views/components/translatable-input.blade.php
```diff
@@ -10,6 +10,10 @@
                 @if ($hint)
                     <x-hearth-hint :for="$name">{{ $hint }}</x-hearth-hint>
                 @endif
+                @isset($interpretationName)
+                    <x-interpretation class="interpretation--start" name="{{ __($interpretationName, [], 'en') }}"
+                        namespace="{{ $interpretationNameSpace }}" />
+                @endisset
```
jobara: As mentioned in the [channel](https://matrix.to/#/!cTDzVKxAmJRmfbSdHL:matrix.org/$FXBZl5FfhzIClsL4ZraE2nFThRch6WDIWXv3Z1x24JY?via=matrix.org), you'll probably need to do the following here: 

```suggestion
                @isset($interpretationName)
                    @isset($interpretationNameSpace)
                        <x-interpretation class="interpretation--start" name="{{ __($interpretationName, [], 'en') }}"
                        namespace="{{ $interpretationNameSpace }}" />
                    @else
                        <x-interpretation class="interpretation--start" name="{{ __($interpretationName, [], 'en') }}" />
                    @endisset
                @endisset
```

This is because blade directives don't work in blade components. 

See: https://github.com/laravel/framework/issues/32045

jobara (changes requested):

jobara:
@chosww the `<x-interpretation>` component in for-individuals.blade.php on line 22 for "What you can do on this website" needs to be updated to be below the text "You can choose how you would like to take part:".

resources/views/individuals/partials/progress.blade.php
```diff
@@ -31,6 +38,13 @@
                 <p id="cannot-publish-explanation">
                     {{ __('You must attend an orientation session and fill in all the required information before you can publish your page.') }}
                 </p>
+                @can('publish', $individual)
+                    <x-interpretation class="interpretation--start" name="{{ __('Publish page', [], 'en') }}"
+                        namespace="publish_page" />
+                @else
+                    <x-interpretation class="interpretation--start" name="{{ __('Publish page (not available)', [], 'en') }}"
+                        namespace="publish_page-disabled" />
+                @endcan
```
jobara: This is nested in the `@cannont` so the `<x-interpretation>` component for the publish button is never reachable. Rather than having a `@can` directive here the publish button interpretation should be moved up to where the publish button is actually added in the template on line 33. The interpretation for the when the it isn't publishable can stay where it is.

resources/views/individuals/edit/about-you.blade.php
```diff
@@ -12,78 +12,106 @@
...
 
             <fieldset>
                 <legend>{{ __('Where do you live?') }}</legend>
+                <x-interpretation class="interpretation--start" name="{{ __('Where do you live?', [], 'en') }}" />
 
                 <div class="field @error('region') field--error @enderror">
                     <x-hearth-label
                         for="region"><x-required>{{ __('Province or territory') }}</x-required></x-hearth-label>
+                    <x-interpretation class="interpretation--start"
+                        name="{{ __('Province or territory (required)', [], 'en') }}" namespace="province_territory" />
```
jobara: Should remove "(required)" from the interpretation name. Also did we decide to update all the form inputs to have the "required" and "optional" in the namespace. I think that was the case, and we should update all of the form input related ones. Please correct me if I'm mistaken.

resources/views/individuals/edit/about-you.blade.php
```diff
@@ -96,21 +124,37 @@
                 @endforeach
             </fieldset>
 
+            <hr class="divider--thick">
+
```
jobara, line 132: Should this be thin divider?

resources/views/individuals/edit/about-you.blade.php
```diff
@@ -12,78 +12,106 @@
...
                     </legend>
+                    <x-interpretation class="interpretation--start"
+                        name="{{ __('How can you help a regulated organization? (required)', [], 'en') }}" />
                     <x-hearth-checkboxes name="consulting_services" :options="$consultingServices" :checked="old('consulting_services', $individual->consulting_services ?? [])"
                         hinted="consulting_services-hint" required />
                     <x-hearth-error for="consulting_services" />
                 </fieldset>
             @endif
 
+            <hr class="divider--thick">
```
jobara, line 111: Should this be thin divider?

resources/views/individuals/edit/experiences.blade.php
```diff
@@ -23,21 +24,23 @@
                             ' ' .
                             __('(optional)')" :shortLabel="__('lived experiences')" :hint="__(
                             'Feel free to self-identify your experiences of disability, if you feel it is relevant to your work.',
-                        )" />
+                        )" :interpretationName="'What are your lived experiences of disability or other intersectional identities? (optional)'" />
```
jobara: As mentioned yesterday you can write this and similar cases like the following:

```suggestion
                        )" interpretationName="What are your lived experiences of disability or other intersectional identities? (optional)" />
```

The ":" at the front of the attribute name basically tells it to interpret it as a php; however since this is just a string value you can just pass in the string directly as a normal attribute.

resources/views/individuals/edit/experiences.blade.php
```diff
@@ -23,21 +24,23 @@
...
                     <x-hearth-error for="lived_experience" />
                 </div>
 
                 {{-- Upload a file --}}
             </fieldset>
 
+            <hr class="divider--thick">
```
jobara, line 36: Should this have been a thin divider? Please review all dividers.

resources/views/individuals/partials/progress.blade.php
```diff
@@ -31,6 +38,13 @@
                 <p id="cannot-publish-explanation">
                     {{ __('You must attend an orientation session and fill in all the required information before you can publish your page.') }}
                 </p>
+                @can('publish', $individual)
+                    <x-interpretation class="interpretation--start" name="{{ __('Publish page', [], 'en') }}"
+                        namespace="publish_page" />
+                @else
+                    <x-interpretation class="interpretation--start" name="{{ __('Publish page', [], 'en') }}"
+                        namespace="publish_page-disabled" />
+                @endcan
```
jobara: This still needs to be updated. The publish button is inside of the `@cannot` from line 37. This publish interpretation should be moved up to after line 33. Then you can also remove the `@can` from here.

jobara:
@chosww could you please also re-run `composer localize` after making the updates?

resources/views/individuals/partials/progress.blade.php
```diff
@@ -31,20 +31,17 @@
             <p class="stack">
                 @can('publish', $individual)
                     <button class="secondary" name="preview" value="1">{{ __('Preview page') }}</button>
+                    <x-interpretation class="interpretation--start" name="{{ __('Publish page', [], 'en') }}"
+                        namespace="publish_page" />
```
jobara: Sorry my fault this is the preview button, which incidentally we don't have a sign language video for. The publish interpretation does need to go below line 38 in a new `@can` block. Although things are starting to get muddled here with some many `@can` and `@cannot` blocks. I'm thinking it'd probably be easier to have just one `@can`\`@else` set where the button and interpretation in the first half and a button with the necessary aria attributes, can't publish explanation and related interpretation.

resources/views/individuals/edit/experiences.blade.php
```diff
@@ -52,7 +52,7 @@
             </fieldset>
 
             <fieldset class="stack">
-                <legend>{{ __('Relevant experiences') }} {{ __('(optional)') }}</legend>
+                <legend class="h4">{{ __('Relevant experiences') }} {{ __('(optional)') }}</legend>
```
jobara, line 57: I think there's a thick divider missing between Skills and strengths and relevant experiences.

jobara (approved):

PR #1497, opened 2023-01-25
ci: create seeders/replication for data added/edited via the admin interfaces (resolves #1483)

### LINKED ISSUE #1483: Create seeders/replication for data added/edited via the admin interfaces

The admin panel has access to add/edit model data (database entries) for some of the models. These need to be able to be captured and restored after a redeploy. In particular this is needed as we test staging for the sign language videos. In the future this will be needed to populate information added to production to the staging and development environments.

In the short term, the sign language videos added to the staging site needs to be restored on deploy.

In the long term, the data added via the admin panel in production needs to be replicated on staging. 

An additional consideration is that the method should be useable to replicate/seed the data into development and local environments. 

**Additional context**

The models that can be edited/created via the admin panel are defined in the [Filament/Resources](https://github.com/accessibility-exchange/platform/tree/dev/app/Filament/Resources) directory. At the moment this includes:

- Identity
- Interpretation
- ResourceCollection
- Resource
- Topic

In the future it will also include models related to training.

marvinroman:
Resolves #1483 

- Creates new filesystem disk on permanent storage `storage/app/seeds`.
- Created new artisan command `db:seed:backup` which you can pass a table to backup the tables as a JSON file into `storage/app/seeds`.
- Updated seeders for the tables to first try and pull from the stored file if it exists in order to seed the table.
- Added new seeder for the **interpretations** table.
- Add cache/clear for event, view, route, and config.

## Tests

- [x] Pull the following tables from staging identities, interpretations, resource_collections, resources, topics.
- [x] Deploy those tables to local mysql deployment.
- [x] Run the command `php artisan db:seed:backup` for each table and make sure that the table file is stored correctly.
- [x] Run the command `php artisan db:seed --class=` for each of the filament table seeders.
- [x] Check tables values before and after seeder is run.

### DIFF
--- a/app/Console/Commands/DatabaseRefresh.php
+++ b/app/Console/Commands/DatabaseRefresh.php
@@ -3,6 +3,7 @@
 namespace App\Console\Commands;
 
 use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
 
 class DatabaseRefresh extends Command
 {
@@ -11,7 +12,8 @@ class DatabaseRefresh extends Command
      *
      * @var string
      */
-    protected $signature = 'db:refresh';
+    protected $signature = 'db:refresh
+                            {--backup : Whether to backup first}?';
 
     /**
      * The console command description.
@@ -27,10 +29,35 @@ class DatabaseRefresh extends Command
      */
     public function handle()
     {
+        $options = $this->options();
+
         // Don't run command in production or testing environment
         if (in_array(config('app.env'), ['testing', 'production']) !== true) {
+
+            // whether to backup seeds first
+            if ($options['backup']) {
+                $this->call('db:seed:backup', ['--all' => true]);
+            }
+
             $this->call('migrate:fresh', ['--force' => true]);
             $this->call('db:seed', ['--class' => 'DevSeeder', '--force' => true]);
+
+            print("Truncating filament tables:\r\n");
+
+            $truncate_tables = array("identities", "interpretations", "resource_collections", "resources", "topics");
+            foreach($truncate_tables as $table) {
+                printf("Truncating values in table %s.\r\n", $table);
+                DB::table($table)->delete();
+            }
+
+            print("\r\n");
+            print("Running filament seeders:\r\n");
+
+            $seeder_classes = array("IdentitySeeder", "InterpretationSeeder", "ResourceCollectionSeeder", "ResourceSeeder", "TopicSeeder");
+            foreach($seeder_classes as $seeder_class) {
+                printf("Running seeder %s.", $seeder_class);
+                $this->call('db:seed', ['--class' => $seeder_class]);
+            }
         }
 
         return 0;
--- a/app/Console/Commands/DatabaseRefresh.php
+++ b/app/Console/Commands/DatabaseRefresh.php
@@ -41,23 +41,6 @@ class DatabaseRefresh extends Command
 
             $this->call('migrate:fresh', ['--force' => true]);
             $this->call('db:seed', ['--class' => 'DevSeeder', '--force' => true]);
-
-            print("Truncating filament tables:\r\n");
-
-            $truncate_tables = array("identities", "interpretations", "resource_collections", "resources", "topics");
-            foreach($truncate_tables as $table) {
-                printf("Truncating values in table %s.\r\n", $table);
-                DB::table($table)->delete();
-            }
-
-            print("\r\n");
-            print("Running filament seeders:\r\n");
-
-            $seeder_classes = array("IdentitySeeder", "InterpretationSeeder", "ResourceCollectionSeeder", "ResourceSeeder", "TopicSeeder");
-            foreach($seeder_classes as $seeder_class) {
-                printf("Running seeder %s.", $seeder_class);
-                $this->call('db:seed', ['--class' => $seeder_class]);
-            }
         }
 
         return 0;
--- a/app/Console/Commands/SeederBackupData.php
+++ b/app/Console/Commands/SeederBackupData.php
@@ -17,7 +17,8 @@ class SeederBackupData extends Command
                             {--a|all : Whether to run through all available backups/restores in config}?
                             {--remove : Remove backed up files}?
                             {--restore : Restore the filament table}?
-                            {--t|table=* : Create/remove specific table file}?';
+                            {--t|table=* : Create/remove specific table file}?
+                            {--truncate : Truncate the table before restoring}';
 
     /**
      * The console command description.
@@ -36,7 +37,7 @@ class SeederBackupData extends Command
         $options = $this->options();
 
         // option to use environment to restore or backup to different environment files
-        if (isset($options['env']) && in_array($options['env'], ['production', 'staging', 'local']) === true) {
+        if (isset($options['env']) && in_array($options['env'], config('backup.filament_seeders.environments')) === true) {
             $environment = $options['env'];
         } else {
             $environment = config('app.env');
@@ -76,6 +77,7 @@ class SeederBackupData extends Command
                 $seeder_classes = config('backup.filament_seeders.classes');
 
                 config(['seeder.environment' => $environment]);
+                config(['seeder.truncate' => $options['truncate']]);
 
                 // run through all the seeder classes
                 // else provide a choice of which to restore
@@ -99,6 +101,7 @@ class SeederBackupData extends Command
                 }
 
                 config(['seeder.environment' => null]);
+                config(['seeder.truncate' => false]);
                 return 0;
             }
 
--- a/app/Console/Commands/SeederBackupData.php
+++ b/app/Console/Commands/SeederBackupData.php
@@ -17,7 +17,7 @@ class SeederBackupData extends Command
                             {--a|all : Whether to run through all available backups/restores in config}?
                             {--remove : Remove backed up files}?
                             {--restore : Restore the filament table}?
-                            {--t|table= : Create/remove specific table file}?*';
+                            {--t|table=* : Create/remove specific table file}?';
 
     /**
      * The console command description.
@@ -42,12 +42,15 @@ class SeederBackupData extends Command
             // if the table option is set only remove that tables file
             // else remove all found in the config
             if (isset($options['table'])) {
-                if (in_array($options['table'], $available_tables)) {
-                    Storage::disk('seeds')->delete($options['table'] . ".json");
-                } else {
-                    printf("You have might have misspelled the tablename.\r\nAvailable tables: %s\r\n", implode(', ', $available_tables));
-                    return 1;
+                foreach($options['table'] as $table) {
+                    if (in_array($table, $available_tables)) {
+                        Storage::disk('seeds')->delete($table . ".json");
+                    } else {
+                        printf("You have might have misspelled the tablename.\r\nTable %s not found.\r\nAvailable tables: %s\r\n", $table, implode(', ', $available_tables));
+                        return 1;
+                    }
                 }
+                return 0;
             } else {
                 foreach($available_tables as $table) {
                     Storage::disk('seeds')->delete($table . ".json");
@@ -96,14 +99,16 @@ class SeederBackupData extends Command
 
         // if table is set the default will be to backup the table
         } elseif (isset($options['table'])) {
-            if (in_array($options['table'], $available_tables)) {
-                printf("Backing up table seeder %s\r\n", $options['table']);
-                Storage::disk('seeds')->put($options['table'] . ".json", DB::table($options['table'])->get()->toJson());
-                return 0;
-            } else {
-                printf("You have might have misspelled the tablename.\r\nAvailable tables: %s\r\n", implode(', ', $available_tables));
-                return 1;
+            foreach($options['table'] as $table) {
+                if (in_array($table, $available_tables)) {
+                    printf("Backing up table seeder %s\r\n", $table);
+                    Storage::disk('seeds')->put($table . ".json", DB::table($table)->get()->toJson());
+                } else {
+                    printf("You have might have misspelled the tablename.\r\nTable %s not found.\r\nAvailable tables: %s\r\n", $table, implode(', ', $available_tables));
+                    return 1;
+                }
             }
+            return 0;
         } else {
             printf("Must use at least one option.\r\nTry php artisan help db:seed:backup\r\n");
             return 1;
--- a/app/Console/Commands/SeederBackupData.php
+++ b/app/Console/Commands/SeederBackupData.php
@@ -35,16 +35,24 @@ class SeederBackupData extends Command
     {
         $options = $this->options();
 
+        // option to use environment to restore or backup to different environment files
+        if (isset($options['env']) && in_array($options['env'], ['production', 'staging', 'local']) === true) {
+            $environment = $options['env'];
+        } else {
+            $environment = config('app.env');
+        }
+
         $available_tables = config('backup.filament_seeders.tables');
 
         // removes backuped up files
         if ($options['remove']) {
             // if the table option is set only remove that tables file
             // else remove all found in the config
-            if (isset($options['table'])) {
+            if (count($options['table']) > 0) {
                 foreach($options['table'] as $table) {
                     if (in_array($table, $available_tables)) {
-                        Storage::disk('seeds')->delete($table . ".json");
+                        printf("Deleting table: %s, for environment: %s\r\n", $table, $environment);
+                        Storage::disk('seeds')->delete(sprintf("%s.%s.json", $table, $environment));
                     } else {
                         printf("You have might have misspelled the tablename.\r\nTable %s not found.\r\nAvailable tables: %s\r\n", $table, implode(', ', $available_tables));
                         return 1;
@@ -53,7 +61,8 @@ class SeederBackupData extends Command
                 return 0;
             } else {
                 foreach($available_tables as $table) {
-                    Storage::disk('seeds')->delete($table . ".json");
+                    printf("Deleting table: %s, for environment: %s\r\n", $table, $environment);
+                    Storage::disk('seeds')->delete(sprintf("%s.%s.json", $table, $environment));
                 }
             }
             return 0;
@@ -66,11 +75,13 @@ class SeederBackupData extends Command
             } else {
                 $seeder_classes = config('backup.filament_seeders.classes');
 
+                config(['seeder.environment' => $environment]);
+
                 // run through all the seeder classes
                 // else provide a choice of which to restore
                 if ($options['all']) {
                     foreach($seeder_classes as $seeder_class) {
-                        printf("Running seeder %s\r\n", $seeder_class);
+                        printf("Running seeder %s for environment %s\r\n", $seeder_class, $environment);
                         $this->call('db:seed', ['--class' => $seeder_class]);
                     }
                 } else {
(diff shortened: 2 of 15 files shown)


app/Console/Commands/DatabaseRefresh.php
```diff
@@ -27,10 +29,35 @@ class DatabaseRefresh extends Command
...
+            if ($options['backup']) {
+                $this->call('db:seed:backup', ['--all' => true]);
+            }
+
             $this->call('migrate:fresh', ['--force' => true]);
             $this->call('db:seed', ['--class' => 'DevSeeder', '--force' => true]);
+
+            print("Truncating filament tables:\r\n");
+
+            $truncate_tables = array("identities", "interpretations", "resource_collections", "resources", "topics");
```
jobara: Can the array of tables and the array of seeders be specified in a config file? This will make it easier to update and we could also set it up to optionally source them from an environment variable.

database/seeders/IdentitySeeder.php
```diff
@@ -4,221 +4,235 @@
...
-                    'fr' => isset($identity['description']) ? trans($identity['description'], [], 'fr') : null,
-                ],
-                'clusters' => $identity['clusters'] ?? [],
-            ]);
+            foreach ($identities as $identity) {
+                Identity::firstOrCreate([
+                    'name' => json_decode($identity['name'], true),
+                    'description' => json_decode($identity['description'], true),
+                    'clusters' => json_decode($identity['clusters'], true),
+                ]);
```
jobara, line 36: I'm not sure exactly but could this have been written as?
```suggestion
                Identity::firstOrCreate(json_decode($identity, true));
```

database/seeders/ResourceSeeder.php
```diff
@@ -18,63 +19,84 @@ class ResourceSeeder extends Seeder
...
+                    'author' => $resource['author'] ?? '',
+                    'summary' => $resource['summary'] ?? '',
+                    'url' => $resource['url'],
+                    'phases' => $resource['phases'] ?? [],
+                ]);
 
-            if (isset($resource['impacts'])) {
-                $item->impacts()->attach($resource['impacts']);
-            }
+                // TODO how do we save and do the connected values?
```
jobara: Is this TODO still relevant, as it looks like you've addressed it below?

app/Console/Commands/DatabaseRefresh.php
```diff
@@ -27,10 +29,35 @@ class DatabaseRefresh extends Command
...
+            }
+
+            print("\r\n");
+            print("Running filament seeders:\r\n");
+
+            $seeder_classes = array("IdentitySeeder", "InterpretationSeeder", "ResourceCollectionSeeder", "ResourceSeeder", "TopicSeeder");
+            foreach($seeder_classes as $seeder_class) {
+                printf("Running seeder %s.", $seeder_class);
+                $this->call('db:seed', ['--class' => $seeder_class]);
+            }
```
jobara: I'm not sure this step is required. The Seeders are written in a way that they will look for a backup first and if not present use some defaults. With the exception of the InterpretationSeeder, all of the updated seeders are run as part of the DevSeeder (some via the DatabaseSeeder that the DevSeeder runs). I think we could just add the InterpretationSeeder to the DevSeeder and all this can be removed.

app/Console/Commands/SeederBackupData.php
```diff
@@ -0,0 +1,55 @@
...
+class SeederBackupData extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'db:seed:backup
+                            {--all : Whether to backup all available tables}?
+                            {--table= : Backup specific table}?*';
```
jobara: We should probably also add a command to clear the backup. This would be helpful for local development so we would haven't to manually delete the seeder backups.

database/seeders/IdentitySeeder.php
```diff
@@ -4,221 +4,235 @@
...
-                'name' => __('Métis'),
-                'clusters' => ['indigenous'],
-            ],
-            [
-                'name' => __('Supporters'),
-                'clusters' => ['lived-experience', 'reachable-when-mixed'],
-            ],
-        ];
+        // check if there is a JSON file that has stored data for the seeder
+        if (Storage::disk('seeds')->exists('identities.json')) {
```
jobara: I guess we might also want to add a conditional check here that we aren't in testing. Some of the seeders may be used when running tests and we wouldn't want the backup data populated there. Another option would be to split up the seeders so the default values (the original ones) are kept clean and called by the new seeders if there aren't backups. Then these original ones will be used in testing. Although we'd have to check if the DevSeeder is used in tests, which could make things a bit more complicated to sort out.

app/Console/Commands/SeederBackupData.php
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            return 0;
+        } elseif (isset($options['table']) && in_array($options['table'], $available_tables)) {
+            Storage::disk('seeds')->put($options['table'] . ".json", DB::table($options['table'])->get()->toJson());
+
+            return 0;
+        } else {
+
+            printf("Must use option --all or --table=. See --help for more details.\r\n");
```
jobara: I hit this error when I mistyped the table name "topics" as "Topics". We should update the error message to include that the table name might be wrong, or have a separate error message for that case.

database/seeders/ResourceSeeder.php
```diff
@@ -18,63 +19,84 @@ class ResourceSeeder extends Seeder
...
+                $item = Resource::firstOrCreate([
+                    'title' => json_decode($resource['title'], true),
+                    'slug' => json_decode($resource['slug'], true),
+                    'summary' => json_decode($resource['summary'], true),
+                    'url' => json_decode($resource['url'], true),
+                    'formats' => json_decode($resource['formats'], true),
+                    'phases' => json_decode($resource['phases'], true),
+                    'author' => json_decode($resource['author'], true),
+                    'content_type_id' => $resource['content_type_id'],
+                    'organization_id' => $resource['organization_id'],
```
jobara, line 65: I tried adding a new resource that contained a reference to an organization_id. We don't backup up the organizations so this triggered an SQL error. 

```error
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`accessibilityexchange`.`resources`, CONSTRAINT `resources_organization_id_foreign` FOREIGN KEY (`organization_id`) REFERENCES `organizations` (`id`) ON DELETE CASCADE) (SQL: insert into `resources` (`title`, `slug`, `summary`, `url`, `formats`, `phases`, `author`, `content_type_id`, `organization_id`, `updated_at`, `created_at`) values ({"en":"New Resource","fr":null}, {"en":"new-resource","fr":"new-resource","asl":"new-resource","lsq":"new-resource"}, {"en":"Resource Summary","fr":null}, {"en":"http:\/\/example.com","lsq":null}, ?, ["design"], {"en":"Comm Org","fr":null}, 2, 1, 2023-01-26 13:48:06, 2023-01-26 13:48:06)/*url='%2F'*/;)
```

I'm leaning towards pulling the backup of the Resources and Resource Collections. The other option would be to remove the organization_id when recreating but the could cause a problem if the author name wasn't provided and also could cause some confusion as data would go missing between deploys of staging that wouldn't be expected to be missing.

jobara (changes requested):

jobara:
@marvinroman would you have time to write up some docs about these new features to add to the Readme?

jobara:
@marvinroman I'm also wondering what will happen when any of the tables change/migrate. Will the restoring from the backup work because they may not line up properly anymore?

database/seeders/IdentitySeeder.php
marvinroman: No. `$identity` is already an object but the property values are strings that need decoded.

app/Console/Commands/DatabaseRefresh.php
marvinroman: It's an array so probably better as just as configuration variables.

I will push changes to the PR that will have the following additional configurations in **backup.php**.

```php
    'filament_seeders' => [
        'tables' => [
            'identities',
            'interpretations',
            'resource_collections',
            'resources',
            'topics',
        ],
        'classes' => [
            'IdentitySeeder',
            'InterpretationSeeder',
            'ResourceCollectionSeeder',
            'ResourceSeeder',
            'TopicSeeder',
        ]
    ],
```

database/seeders/ResourceSeeder.php
marvinroman: Will be resolved in next commits. We already discussed this TODO.

app/Console/Commands/DatabaseRefresh.php
jobara: The InterpretationSeeder would probably also need to be added to the ProductionSeeder.

marvinroman: The classes will be part of the config so later we can place somewhere to restore all these tables with a single command.

I will remove this and add the **InterpretationSeeder** to the **DevSeeder**.

app/Console/Commands/SeederBackupData.php
marvinroman: Adding both `--remove` & `--restore` options.

app/Console/Commands/SeederBackupData.php
marvinroman: This will be the error if they used the tablename option but wasn't in the list of available tables.
```bash
You have might have misspelled the tablename.
Available tables: identities, interpretations, resource_collections, resources, topics
```

database/seeders/ResourceSeeder.php
marvinroman: Here is another option. We just skip those lines and show the error so it can be fixed if the user is restoring at the command line. Here is the testing output that I ran a test on.

```bash

 Which class would you like to restore? [InterpretationSeeder]:
  [0] IdentitySeeder
  [1] InterpretationSeeder
  [2] ResourceCollectionSeeder
  [3] ResourceSeeder
  [4] TopicSeeder
 > 3

Running seeder ResourceSeeder
   INFO  Seeding database.  

Skipping insert because content type with id 1 is missing 
Skipping insert because content type with id 1 is missing 
Skipping insert because content type with id 1 is missing 
```

marvinroman: Let me know if you want me to keep this? We can also remove these from the configuration for tables and classes in the short term and come back for more solid solutions in the future.

database/seeders/IdentitySeeder.php
marvinroman: Added this check for all the seeders sections that I added to prevent them from running in testing or production.

```php
in_array(config('app.env'), ['testing', 'production']) !== true && 
```

marvinroman:
> 

It would

marvinroman:
> I'm also wondering what will happen when any of the tables change/migrate. Will the restoring from the backup work because they may not line up properly anymore?

It really depends on the change.

- If a column was deleted we would have to delete it from the seeder class.  
- If a column type was changed we might have to change how it's added.
- If a column is added we will want to add the the seeder, but make sure it fails gracefully if it's not yet in the JSON file.

jobara:
@marvinroman sorry, I was off work for a couple of days. I'm noticing this is closed now. Was that intentional?

jobara:
> * If a column was deleted we would have to delete it from the seeder class.

This seems to be the easiest and we just need to update the seeder when we make our database migration.

> * If a column type was changed we might have to change how it's added.

I'm guessing here it would be a "one" time change, so we'd have to have some logic that accounted backup being in either the old or new state. I suppose in a future release we'd update this again to only work for the new state. Is that correct?

> * If a column is added we will want to add the the seeder, but make sure it fails gracefully if it's not yet in the JSON file.

This might be a problem because our old data would basically get trashed at this point. That is, it wouldn't be restored. Is that correct? Any thoughts?

I can't think of a way to do this, but what we kind of want to do is run the new migrations on the old database before backing it up.

app/Console/Kernel.php
```diff
@@ -24,14 +24,14 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        $schedule->command('db:refresh') // use custom command to make sure that the commands are chained
+        $schedule->command('db:refresh --backup') // use custom command to make sure that the commands are chained
             ->daily() // Run daily at midnight
             ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
```
jobara, line 29: I guess after we do a real production release, we'll change this to run the backup on Production so that we can replicate the data to staging. Is that correct?

database/seeders/ResourceSeeder.php
jobara: I guess that's fine for this case. So we restore only the ones that don't have missing foreign keys?

jobara (changes requested):

jobara:
@marvinroman I believe the InterpretationSeeder also needs to be added to the ProductionSeeder because that is the one that gets run on staging. We will probably also need to run it initially on staging but I guess it won't really matter that the interpretationSeeder is there, but we will need to manually migrate the Interpretations table over to production when we do that first production release.

Also thanks for the detailed docs.

jobara:
@marvinroman noticed an issue with the resources and resource collections. It seems that after restoring from a backup, the resources collections do not have their resources attached.

jobara:
@marvinroman also noticed that `--remove` only seemed to work if I specified an explicit table name, wasn't able to remove all. Nothing reported and the files were not removed when calling `php artisan db:seed:backup --remove` and `php artisan db:seed:backup --remove --all`. 

Also didn't get any error or anything when running `php artisan db:seed:backup`. I'm guessing it requires the `--all` or specific table name flags.

database/seeders/ResourceSeeder.php
marvinroman: Yes we just make each seeder robust to missing foreign keys if they have them.

app/Console/Kernel.php
marvinroman: Yeah. I expected it would run on both. 
We can have a shared directory between both production and sandbox.
I can add in a command option to allow to pull from different locations?
Example:
Backups occur to `/data/[environment]/seeds` by default. 
If you add `--path=/data/production/seeds` while in sandbox it would restore from production seeds.

marvinroman:
>

marvinroman:
> @marvinroman I believe the InterpretationSeeder also needs to be added to the ProductionSeeder because that is the one that gets run on staging. We will probably also need to run it initially on staging but I guess it won't really matter that the interpretationSeeder is there, but we will need to manually migrate the Interpretations table over to production when we do that first production release.
> 
> Also thanks for the detailed docs.

This update will be part of my next commit.

marvinroman:
> This might be a problem because our old data would basically get trashed at this point. That is, it wouldn't be restored. Is that correct? Any thoughts?

I think we can take these cases as they come up and find a strategy that works for it. I don't think we are going to solve for all cases. In some cases it would make sense to work a data upgrade into the migration and then run the backup after the migration.

marvinroman:
> @marvinroman also noticed that --remove only seemed to work if I specified an explicit table name, wasn't able to remove all. Nothing reported and the files were not removed when calling php artisan db:seed:backup --remove and php artisan db:seed:backup --remove --all.
> 
> Also didn't get any error or anything when running php artisan db:seed:backup. I'm guessing it requires the --all or specific table name flags.

The table check wasn't working after the change to getting options instead of direct parameters. This will be fixed in the next commit.

marvinroman:
> @marvinroman noticed an issue with the resources and resource collections. It seems that after restoring from a backup, the resources collections do not have their resources attached.

Maybe you can show me through this one? Or provide a test case? I need more information on how these links are made and how to test.

app/Console/Kernel.php
marvinroman: I think this might be a better solution.
We can just tag the files with an extension that is the environment.
So for example backing up on my local which has **local** set at the environment then the backup files look like this.
```
identities.json.local  interpretations.json.local  resource_collections.json.local  resources.json.local  topics.json.local
```
So by adding `--env=staging` to any of the commands it will change the relevant files that will be handled.

What do you think of this solution?

jobara: If possible can we move the "tag" between the file name and the ".json" extension. So something like "identities.local.json". That way applications (e.g. an IDE) would still know the file is a JSON file.

jobara:
> Maybe you can show me through this one? Or provide a test case? I need more information on how these links are made and how to test.

I pinged you in Slack to see if we can arrange a time to meet today.

database/seeders/TopicSeeder.php
```diff
@@ -4,29 +4,63 @@
...
+                foreach ($topics as $topic) {
+                    Topic::firstOrCreate([
+                        'name->en' => $topic,
+                        'name->fr' => trans($topic, [], 'fr'),
+                    ]);
+                }
+            }
+        } else {
+            $environment = config('app.env');
+            printf("Seeder cannot be run on environment: %s\r\n", $environment);
```
jobara: If we still need this can we change it to something like the following so we know which seeder didn't run.

```suggestion
            printf("%s cannot be run on environment: %s\r\n", self::class, $environment);
```

The same would apply for all other cases.

database/seeders/IdentitySeeder.php
```diff
@@ -4,221 +4,257 @@
...
-                    'fr' => trans($identity['name'], [], 'fr'),
-                ],
-                'description' => [
-                    'en' => $identity['description'] ?? null,
-                    'fr' => isset($identity['description']) ? trans($identity['description'], [], 'fr') : null,
-                ],
-                'clusters' => $identity['clusters'] ?? [],
-            ]);
+        // option to use environment to restore or backup to different environment files
+        if (config('seeder.environment') !== null && in_array(config('seeder.environment'), ['production', 'staging', 'local']) === true) {
```
jobara: The environment array should have come from the config file. Same throughout.

jobara: Where does `config('seeder.environment')` come from. Is that generated by running the `db:seed:backup` command?

database/seeders/IdentitySeeder.php
```diff
@@ -4,221 +4,257 @@
...
+            $environment = config('seeder.environment');
+        } else {
+            $environment = config('app.env');
+        }
+
+        // check if there is a JSON file that has stored data for the seeder
+        if (in_array(config('app.env'), ['testing', 'production']) !== true) {
```
jobara: It seems like this was added to all of the updated seeders, but this prevents the defaults from running in 'testing' and 'production'. However, in most or all cases the default seeders should have been run there. This is what's causing the tests to fail for example.

app/Console/Commands/SeederBackupData.php
```diff
@@ -0,0 +1,134 @@
...
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'db:seed:backup
+                            {--a|all : Whether to run through all available backups/restores in config}?
+                            {--remove : Remove backed up files}?
+                            {--restore : Restore the filament table}?
+                            {--t|table=* : Create/remove specific table file}?
+                            {--truncate : Truncate the table before restoring}';
```
jobara, line 21: it looks like we're missing the entry for `--env`.

jobara (changes requested):

app/Console/Commands/SeederBackupData.php
marvinroman: `--env` is inherited by all commands already.

database/seeders/IdentitySeeder.php
marvinroman: `config('seeder.environment')` is how we pass the environment option that was set in the command to the seeder.

marvinroman: > The environment array should have come from the config file. Same throughout.

This will be in the next commit.

database/seeders/IdentitySeeder.php
marvinroman: Removed from seeder classes.

database/seeders/TopicSeeder.php
marvinroman: We are removing the check altogether.

jobara (approved):

PR #1514, opened 2023-02-01
feat: add orientation sign up links (resolves #1212)

### LINKED ISSUE #1212: Add orientation signup links

Individual, organizational and regulated organizational accounts need to complete orientation.

Their dashboards should provide links to sign up for their respective orientation sessions:

- Individual (Participant): https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=3706%3A98663
- Individual (Consultant/Connector): https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=1784%3A51410
- Organization: https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=2118%3A56197
- Regulated Organization: https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?node-id=1491%3A35377

**Additional context**

We still don't have the actual links.

chosww:
Add orientation sign up links for users - #1212


jobara:
@chosww I believe we also need to add these links when creating their pages. The publish button is disabled with a message about needing to attend an orientation session. If you search for "You must attend an orientation session" you should be able to find the spots. I think you'll just need to convert "orientation session" into a link. You can do this by using either [`Str::markdown()`](https://laravel.com/docs/9.x/helpers#method-str-markdown) or [`Str.inlineMarkdown()`](https://laravel.com/docs/9.x/helpers#method-str-inline-markdown)

For example:

```php
{{ Str.markdown(__('You must attend an [orientation session](:url) and fill in all the required information before you can publish your page.', [url: http://example.com])) }}
```

app/Http/Controllers/ProjectController.php
```diff
@@ -114,12 +114,22 @@ public function show(Project $project): View
 
     public function edit(Project $project): View
     {
+        $orientationLink = '#';
+        if (Auth::user()->context === 'regulated-organization') {
+            $orientationLink = 'https://share.hsforms.com/1gGf9TjhaQ0uaqcnyJfSDlwdfpez';
+        } elseif (Auth::user()->context === 'organization') {
+            $orientationLink = 'https://share.hsforms.com/1sB6UV4gvQlC_0QxQ3q3z1Adfpez';
+        } elseif (Auth::user()->context === 'individual') {
+            $orientationLink = 'https://share.hsforms.com/161eyaBsQS-iv1z0TZLwdQwdfpez';
```
jobara: An alternative way to write this would be to use [match](https://www.php.net/manual/en/control-structures.match.php). If we update this again we should also consider moving this to somewhere that can be reused to so that we can update the links in a single location if needed again.

jobara:
@chosww could you please address the merge conflict. It's with the localization files.

app/Http/Controllers/ProjectController.php
jobara: Actually now that there's a merge conflict can you update this to use the match statement. Also consider moving this to a helper function or something else and making the call to it in the other locations. You could probably also write a test for the logic then. Another option instead of a helper is an Enum. There may be other ways to generalize it too.

resources/views/dashboard/individual.blade.php
```diff
@@ -16,6 +16,12 @@
                 </li>
             @endcan
         @endif
+        @if (!$user->oriented_at)
+            <li>
+                <a
+                    href="https://share.hsforms.com/161eyaBsQS-iv1z0TZLwdQwdfpez">{{ __('Sign up for an orientation session') }}</a>
```
jobara: I think you forgot to use your orientation helper function in many of the spots.

jobara (changes requested):

resources/views/individuals/partials/progress.blade.php
```diff
@@ -39,7 +39,12 @@
                         namespace="publish_page" />
                 @else
                 <p id="cannot-publish-explanation">
-                    {{ __('You must attend an orientation session and fill in all the required information before you can publish your page.') }}
+                    {!! Str::markdown(
+                        __(
+                            'You must attend an [orientation session](:url) and fill in all the required information before you can publish your page.',
+                            ['url' => 'https://share.hsforms.com/161eyaBsQS-iv1z0TZLwdQwdfpez'],
```
jobara: There some more that still need to be updated.

resources/views/dashboard/regulated-organization.blade.php
```diff
@@ -22,6 +22,11 @@
                 @endif
             @endcan
         @endcan
+        @if (!$user->oriented_at)
```
jobara: Need to check the Regulated Organization. Try:
```suggestion
        @if (!$user->regulatedOrganization->oriented_at)
```

resources/views/dashboard/organization.blade.php
```diff
@@ -20,6 +20,11 @@
                 @endif
             @endcan
         @endcan
+        @if (!$user->oriented_at)
```
jobara: This needs to actually check the organization's `oriented_at` property. You can try: 
```suggestion
        @if (!$user-> organization->oriented_at)
```

resources/views/individuals/partials/progress.blade.php
```diff
@@ -39,7 +39,12 @@
                         namespace="publish_page" />
                 @else
                 <p id="cannot-publish-explanation">
-                    {{ __('You must attend an orientation session and fill in all the required information before you can publish your page.') }}
+                    {!! Str::markdown(
+                        __(
+                            'You must attend an [orientation session](:url) and fill in all the required information before you can publish your page.',
+                            ['url' => orientation_link('individual')],
```
jobara: It's better not to hardcode the argument value. First attempt to get it from something passed into the page, for example from the User. If that isn't possible you should get it via the enum (e.g. `orientation_link(App\Enums\UserContext::Individual->value)`). That will minimize the maintenance needed for name changes.

app/helpers.php
```diff
@@ -245,3 +245,15 @@ function contact_information(): string
...
     }
 }
+
+if (! function_exists('orientation_link')) {
+    function orientation_link(string $userType): string
+    {
+        return match ($userType) {
+            'individual' => 'https://share.hsforms.com/161eyaBsQS-iv1z0TZLwdQwdfpez',
+            'organization' => 'https://share.hsforms.com/1sB6UV4gvQlC_0QxQ3q3z1Adfpez',
+            'regulated-organization' => 'https://share.hsforms.com/1gGf9TjhaQ0uaqcnyJfSDlwdfpez',
```
jobara: Can we get the keys from the UserContext enum so that we guarantee that it matches?

jobara (changes requested):

resources/views/individuals/partials/progress.blade.php
chosww: Thanks @jobara that makes sense not to use hardcoded value here for the maintenance. I was wondering, inside of the progress.blade for regulated organization, I am using Auth::user()->context, and could we possibly use the same approach here than using enum value?

jobara: @chosww you could probably if that blade file is only ever used for regulated organizations. Otherwise pulling the context from the User would work as well because it should use the enum value internally.

jobara (approved):

PR #1532, opened 2023-02-08
feat: Engagement: If full, show message (resolves #1271)

### LINKED ISSUE #1271: Engagement: If full, show message
When an open call engagement has been filled up, there is nothing to indicate that to users.

In place of the "sign up" button please have the text:
"All participant spots have been filled."

chosww:
Resolves #1271 

Disable signup button and show a message about that when engagements reach their participants capacity limit.

### Prerequisites


resources/views/engagements/show.blade.php
```diff
@@ -59,9 +59,16 @@
...
+                <div class="flex flex-col">
+                    <a class="{{ $engagement->confirmedParticipants->count() >= $engagement->ideal_participants ? 'cta disabled' : 'cta' }} mx-auto"
+                        href="{{ localized_route('engagements.sign-up', $engagement) }}"
+                        tabindex="{{ $engagement->confirmedParticipants->count() >= $engagement->ideal_participants ? '-1' : '0' }}">
+                        @svg('heroicon-o-clipboard-check') {{ __('Sign up') }}
+                    </a>
+                    @if ($engagement->confirmedParticipants->count() >= $engagement->ideal_participants)
+                        <p>{{ __('All participant spots have been filled.') }}</p>
+                    @endif
+                </div>
```
jobara: The disabled state should use the `@ariaDisabled` blade directive that is defined in this project. See how it is applied to buttons like the [disabled publish button](https://github.com/accessibility-exchange/platform/blob/019fa76639be469203461544f9bfb099af3dcf81/resources/views/engagements/edit.blade.php#L296).

resources/css/components/_link.css
```diff
@@ -84,6 +84,14 @@ a.cta {
     width: auto;
 }
 
+a.cta.disabled {
+    background: transparent;
+    border: var(--border-2) dashed;
+    border-color: var(--cta-border-hover);
+    color: var(--cta-color-hover);
+    pointer-events: none;
+}
```
jobara: Styling for the disabled state should be handled similar to the disabled buttons. See: [_button.css](https://github.com/accessibility-exchange/platform/blob/019fa76639be469203461544f9bfb099af3dcf81/resources/css/components/_button.css#L258-L265)

resources/views/engagements/show.blade.php
jobara: Another issue is that the disabled state will never show up. The `@can` directive used on line 61 will prevent the sign up button from showing if the capacity has been reached. 

If we still need to show a disabled "sign up" link, we may need to do a check for all the following to decide if it is displayed:

- cannot "join"
- cannot "participate" (see line 74 below)
- engagement confirmed count is greater than or equal to the ideal number of participants.

jobara (changes requested):

resources/views/engagements/show.blade.php
chosww: I realized that join policy already includes check for the participants count; I updated the policy to not to check the participants counts, because we want to have all the other rules specified for the join. Also, I don't think we need to consider case of cannot participate.

resources/views/engagements/show.blade.php
```diff
@@ -58,10 +58,16 @@
...
+            @can('requestToJoin', $engagement)
+                <div class="flex flex-col">
+                    <a class="cta mx-auto" href="{{ localized_route('engagements.sign-up', $engagement) }}"
+                        @if ($engagement->confirmedParticipants->count() >= $engagement->ideal_participants) @ariaDisabled @endif>
+                        @svg('heroicon-o-clipboard-check') {{ __('Sign up') }}
+                    </a>
+                    @if ($engagement->confirmedParticipants->count() >= $engagement->ideal_participants)
+                        <p>{{ __('All participant spots have been filled.') }}</p>
+                    @endif
+                </div>
```
jobara: The description text should be linked to the button as an `aria-describedby`. For example:

```suggestion
            @can('requestToJoin', $engagement)
                <div class="flex flex-col">
                    <a class="cta mx-auto" href="{{ localized_route('engagements.sign-up', $engagement) }}"
                        @if ($engagement->confirmedParticipants->count() >= $engagement->ideal_participants) @ariaDisabled aria-describedby="engagement-full-explanation" @endif>
                        @svg('heroicon-o-clipboard-check') {{ __('Sign up') }}
                    </a>
                    @if ($engagement->confirmedParticipants->count() >= $engagement->ideal_participants)
                        <p id="engagement-full-explanation">{{ __('All participant spots have been filled.') }}</p>
                    @endif
                </div>
```

jobara (changes requested):

jobara (approved):

Issue #1161, opened 2022-10-12
Users contracted for project loses ability to view project they are contracted for in their dashboard


chosww:
**Prerequisites**

- [x] I've [searched](https://github.com/accessibility-exchange/platform/issues) for existing issues

**Describe the bug**
Removing community connector or Accessibility Consultant role from an individual who is contracted for projects, the individual loses ability to view "Project I'm contracted for" option from their dashboard.

**To Reproduce**
1. Log into an account that is contracted for a project.
2. Update their role to not have either community connector nor accessibility consultant.
3. Go to Dashboard
4. See that the "Project I'm contracted for" option is not present.

**Expected behavior**
User should still be able to see "Project I'm contracted for" option on their dashboard.

cherylhjli:
This should also apply to the tabs on the projects page as well.

michelled:
Completed in the above pull request

PR #1535, opened 2023-02-13
fix: separate the projects into different tabs based on users roles (resolves #1469)

### LINKED ISSUE #1469: Separate the projects into 3 tabs based on users roles

This is a change requested in #1419 -- "Clarify the labels for “Projects I am participating” and “Projects I am contracted” ( we have not decided on the final phrasing yet) ex for Projects I am contracted: projects I am supporting, organizations I am involved with"

Separate the projects into 3 tabs instead of 2, and naming them:
"Involved as a Consultation Participant"
"Involved as an Accessibility Consultant"
"Involved as a Community Connector"
If the user only has 1 role, don't show any tabs.
If the user only has 1 role currently but previously had another role that still has an active project they're working on, show that tab. (https://github.com/accessibility-exchange/platform/issues/1161)

chosww:
Resolves #1469 

Separate the projects into different tabs based on users roles.

### Prerequisites


resources/views/projects/my-projects.blade.php
```diff
@@ -18,54 +18,71 @@
...
-            <div class="center center:wide">
-                <ul class="-mt-4 flex gap-6" role="list">
-                    @if ($user->context === 'organization')
-                        @if ($user->organization->isConsultant() || $user->organization->isConnector())
+        @if ($user->context === 'organization' ||
+            ($user->context === 'individual' && $user->individual->roles && sizeof($user->individual->roles) > 1) ||
+            ($user->indiviudal?->inProgressContractedProjects &&
+                sizeof($user->indiviudal->inProgressContractedProjects) > 0 &&
+                $user->individual?->inProgressParticipatingProjects &&
+                sizeof($user->individual->inProgressParticipatingProjects) > 0))
```
jobara: You check if `$user->individual` is `null` in some cases but not others. I'd assume if you need to check it sometimes, you'd need to check it all the time.

resources/views/projects/my-projects.blade.php
```diff
@@ -18,54 +18,71 @@
...
+                    <ul class="-mt-4 flex gap-6" role="list">
+                        @if ($user->context === 'organization')
+                            {{-- @if ($user->organization->isConsultant() || $user->organization->isConnector())
+                                <li class="w-full">
+                                    <x-nav-link class="inline-flex w-full items-center justify-center border-t-0"
+                                        :href="localized_route('projects.my-projects')" :active="request()->localizedRouteIs('projects.my-projects')">
+                                        {{ __('Involved as an Accessibility Consultant') }}
+                                    </x-nav-link>
+                                </li>
+                            @endif --}}
```
jobara: Please leave a comment about why this is commented out, and provide a link to a GitHub issue preferably.

jobara (changes requested):

jobara:
@chosww #1469 mentions in it's requested solutions a fix for #1161. I don't see that addressed in this PR though, should I be expecting that?

resources/views/projects/my-projects.blade.php
chosww: When user->context is individual, that implies that $user->individual won't be null and using the check inside of sizeof throws an error that says sizeof can't be used on potentially null variables. To clarify, was your comment about adding checks when using  $user->individual with $user->context?

resources/views/projects/my-projects.blade.php
chosww: Sounds good, do we have a GitHub issue for implementing consultant feature?

resources/views/projects/my-projects.blade.php
jobara: Actually my comment was really about the last `||` section. Where you check `$user->indiviudal?->inProgressContractedProjects` is `null` (also there's a typo there). Then also check again on line 25.

Regarding `sizeof`if there is a chance it could still result in `$user->individual` being `null` you'd probably want to do something like `sizeof($user->individual?->roles ?? [])`.

resources/views/projects/my-projects.blade.php
jobara: I couldn't find one. Maybe file one and then ask @cherylhjli to flesh it out a bit. It's something that's probably out of scope for the time being though, but would be good to have enough info that we know what it is about.

resources/views/dashboard/individual.blade.php
```diff
@@ -1,6 +1,8 @@
 <div class="with-sidebar with-sidebar:2/3">
     <x-quick-links>
-        @if ($user->individual->isConnector() || $user->individual->isConsultant())
+        @if ($user->individual->isConnector() ||
+            $user->individual->isConsultant() ||
+            $user->individual->inProgressContractedProjects)
```
jobara: Likely should be the following:

```suggestion
            $user->individual->inProgressContractedProjects()->count())
```

resources/views/dashboard/individual.blade.php
```diff
@@ -22,9 +24,10 @@
             </li>
         @endif
         @can('viewAny', App\Models\Project::class)
-            @if ($user->individual->isParticipant())
+            @if ($user->individual->isParticipant() || $user->individual->inProgressParticipatingProjects)
```
jobara: Likely should be the following:
```suggestion
            @if ($user->individual->isParticipant() || $user->individual->inProgressParticipatingProjects()->count())
```

resources/views/projects/my-projects.blade.php
```diff
@@ -18,54 +18,73 @@
         ($user->context === 'individual' &&
             ($user->individual->isConsultant() || $user->individual->isConnector()) &&
             $user->individual->isParticipant()))
```
jobara: This likely needs to check if the user is participating in projects.

```suggestion
            ($user->individual->isParticipant() || $user->individual->inProgressParticipatingProjects()->count())))
```

resources/views/projects/my-projects.blade.php
```diff
@@ -18,54 +18,73 @@
...
-        <nav class="nav--tabbed" aria-labelledby="projects">
-            <div class="center center:wide">
-                <ul class="-mt-4 flex gap-6" role="list">
-                    @if ($user->context === 'organization')
-                        @if ($user->organization->isConsultant() || $user->organization->isConnector())
+        @if ($user->context === 'organization' ||
+            ($user->context === 'individual' && $user->individual->roles && sizeof($user->individual?->roles ?? []) > 1) ||
```
jobara: `$user->individual->roles && sizeof($user->individual?->roles ?? []) > 1` is a bit redundant. You can just check 

```suggestion
            ($user->context === 'individual' && count($user->individual?->roles ?? []) > 1) ||
```

resources/views/projects/my-projects.blade.php
```diff
@@ -18,54 +18,73 @@
...
+            ($user->individual?->inProgressContractedProjects &&
+                sizeof($user->individual?->inProgressContractedProjects ?? []) > 0 &&
+                $user->individual?->inProgressParticipatingProjects &&
+                sizeof($user->individual?->inProgressParticipatingProjects ?? []) > 0))
```
jobara: This can be written as the following:
```suggestion
            ($user->individual?->inProgressContractedProjects()->count() &&
                $user->individual?->inProgressParticipatingProjects()->count()))
```

jobara (changes requested):

jobara:
Instead of using `sizeof()` please use `count()` or `->count()` if a laravel collection. `sizeof` is just an alias of `count` so for consistency it would be better to use `count()`.

jobara:
If I sign up as a consultant as an individual but remove the role, I can no longer see the project/engagement in the projects page even though I'm still an active participant. The same for an individual as a community connector. Please also check for organizations.

chosww:
> If I sign up as a consultant as an individual but remove the role, I can no longer see the project/engagement in the projects page even though I'm still an active participant. The same for an individual as a community connector. Please also check for organizations.

Thanks for reviewing the PR Justin. I am not understanding the issue you commented, could we meet and talk about it next week?

app/Http/Controllers/UserProjectsController.php
```diff
@@ -17,20 +17,19 @@ public function show(): Response|View
             $projectable->load('projects');
         }
 
-        if ($user->organization) {
-            $projectable = $user->organization;
-            if ($projectable->isConsultant() || $projectable->isConnector()) {
-                $section = 'contracted';
-            } elseif ($projectable->isParticipant()) {
+        if (in_array($user->context, ['individual', 'organization'])) {
+            $userContext = $user->individual ?? $user->organization;
```
jobara: I think you can write this as follows:
```suggestion
            $userContext = $user->{$user->context}
```

app/Http/Controllers/UserProjectsController.php
```diff
@@ -17,20 +17,19 @@ public function show(): Response|View
             $projectable->load('projects');
         }
 
-        if ($user->organization) {
-            $projectable = $user->organization;
-            if ($projectable->isConsultant() || $projectable->isConnector()) {
-                $section = 'contracted';
-            } elseif ($projectable->isParticipant()) {
+        if (in_array($user->context, ['individual', 'organization'])) {
```
jobara: The array values should be sourced from the UserContext enum like:

```suggestion
        if (in_array($user->context, [UserContext::Individual->value, UserContext::Organization->value])) {
```

app/Http/Controllers/UserProjectsController.php
```diff
@@ -46,9 +45,11 @@ public function showContracted(): Response|View
         $user = Auth::user();
 
         if (in_array($user->context, ['individual', 'organization'])) {
-            if ($user->individual && $user->individual->isParticipant() && ($user->individual->isConsultant() || $user->individual->isConnector())) {
+            $userContext = $user->individual ?? $user->organization;
+            if ($userContext && ($userContext->isConsultant() || $userContext->isConnector() || $userContext->inProgressContractedProjects()->count())) {
```
jobara: Looks like this is the second time this conditional is being checked. Makes me wonder if you should store these complex conditionals in variables such as `isParticipant` and `isContracted` or something like that.

app/Http/Controllers/UserProjectsController.php
```diff
@@ -46,9 +45,11 @@ public function showContracted(): Response|View
...
                 return view('projects.my-projects', [
                     'user' => $user,
+                    'projectable' => $user->organization ?? null,
```
jobara: The nullish coalescence is unnecessary here because it will only trigger if the left side is already `null`.
```suggestion
                    'projectable' => $user->organization,
```

app/Http/Controllers/UserProjectsController.php
```diff
@@ -61,12 +62,15 @@ public function showParticipating(): Response|View
...
         $user = Auth::user();
 
-        if ($user->context === 'organization' && $user->organization && $user->organization->isParticipant() && ($user->organization->isConsultant() || $user->organization->isConnector())) {
-            return view('projects.my-projects', [
-                'user' => $user,
-                'projectable' => $user->organization,
-                'section' => 'participating',
-            ]);
+        if (in_array($user->context, ['individual', 'organization'])) {
+            $userContext = $user->individual ?? $user->organization;
```
jobara: See previous comments above.

app/Http/Controllers/UserProjectsController.php
```diff
@@ -61,12 +62,15 @@ public function showParticipating(): Response|View
...
+            if ($userContext && ($userContext->isParticipant() || $userContext->inProgressParticipatingProjects()->count())) {
+                return view('projects.my-projects', [
+                    'user' => $user,
+                    'projectable' => $user->organization ?? null,
```
jobara: See previous comment about nullish coalescence.

app/Http/Controllers/UserProjectsController.php
```diff
@@ -76,14 +80,12 @@ public function showRunning(): Response|View
...
 
-        if ($user->context === 'organization') {
-            if ($user->organization && ($user->organization->isParticipant() || $user->organization->isConsultant() || $user->organization->isConnector())) {
-                return view('projects.my-projects', [
-                    'user' => $user,
-                    'projectable' => $user->organization,
-                    'section' => 'running',
-                ]);
-            }
+        if (in_array($user->context, ['organization', 'regulated-organization'])) {
```
jobara: See comment about sourcing these from the UserContext enum

resources/views/projects/my-projects.blade.php
```diff
@@ -17,55 +17,66 @@
...
             ($user->individual->isConsultant() || $user->individual->isConnector()) &&
-            $user->individual->isParticipant()))
-        <nav class="nav--tabbed" aria-labelledby="projects">
-            <div class="center center:wide">
-                <ul class="-mt-4 flex gap-6" role="list">
-                    @if ($user->context === 'organization')
-                        @if ($user->organization->isConsultant() || $user->organization->isConnector())
+            ($user->individual->isParticipant() || $user->individual->inProgressParticipatingProjects()->count())))
+        @if ($user->context === 'organization' ||
+            ($user->context === 'individual' && count($user->individual?->roles ?? []) > 1) ||
```
jobara: These context comparisons should compare against a value sourced from the UserContext enum

resources/views/projects/my-projects.blade.php
```diff
@@ -17,55 +17,66 @@
...
+            ($user->individual?->inProgressContractedProjects()->count() &&
+                $user->individual?->inProgressParticipatingProjects()->count()))
+            <nav class="nav--tabbed" aria-labelledby="projects">
+                <div class="center center:wide">
+                    <ul class="-mt-4 flex gap-6" role="list">
+                        @if ($user->context === 'organization')
```
jobara: See comment about sourcing from UserContext enum

resources/views/projects/my-projects.blade.php
```diff
@@ -17,55 +17,66 @@
...
                                 <x-nav-link class="inline-flex w-full items-center justify-center border-t-0"
-                                    :href="localized_route('projects.my-projects')" :active="request()->localizedRouteIs('projects.my-projects')">
-                                    {{ __('Projects I am contracted for') }}
+                                    :href="localized_route('projects.my-running-projects')" :active="request()->localizedRouteIs('projects.my-running-projects')">
+                                    {{ __('Projects I am running') }}
                                 </x-nav-link>
                             </li>
                         @endif
-                        @if ($user->organization->isParticipant())
+                        @if ($user->context === 'individual')
```
jobara: See comment about sourcing from UserContext enum

jobara (changes requested):

app/Http/Controllers/UserProjectsController.php
```diff
@@ -61,12 +62,14 @@ public function showParticipating(): Response|View
     {
         $user = Auth::user();
 
-        if ($user->context === 'organization' && $user->organization && $user->organization->isParticipant() && ($user->organization->isConsultant() || $user->organization->isConnector())) {
-            return view('projects.my-projects', [
-                'user' => $user,
-                'projectable' => $user->organization,
-                'section' => 'participating',
-            ]);
+        if (in_array($user->context, ['individual', 'organization'])) {
```
jobara: Missed sourcing the array values from the UserContext enum. This is the third time this conditional has been checked. You may want to move it into a separate function. e.g. a shared function in this controller, a helper function, or defined in the user model.

app/Http/Controllers/UserProjectsController.php
jobara: The array values should be sources from the UserContext enum.

jobara (changes requested):

jobara:
@chosww looks like this doesn't address the issue from #1161. I added an individual as a community connector for one engagement and a participant for another. If I change the roles, you can't see the projects for corresponding for the role(s) that were removed (same with dashboard link). These should still be visible while they are in active projects.

jobara (approved):

PR #1609, opened 2023-03-24
Ci/remove linking to shared permanent storage cache

### LINKED ISSUE #1596: Class "Spatie\SqlCommenter\SqlCommenterServiceProvider" not found
**`Error`**

Class "Spatie\SqlCommenter\SqlCommenterServiceProvider" not found

:dart: Illuminate\Foundation\Application::resolveProvider
:page_facing_up: /app/vendor/laravel/framework/src/Illuminate/Foundation/Application.php

```php
748      public function resolveProvider($provider)
749      {
750*         return new $provider($this);
751      }
752  
```
[Open in Flare](https://flareapp.io/errors/3468651/latest)

marvinroman:
* Removes linking of local pod file caching link from permanent/shared storage in entrypoint.
* Moves artisan commands that create files within cache directory from `deploy:global` to `deploy:local`.
* Added artisan command `livewire:discover` to create livewire components cache, a file I found that is created in cache but no generated by other commands.

This PR was branched from **staging** and being pushed back to **staging** since it is build related.

### DIFF
--- a/app/Console/Commands/DeployLocal.php
+++ b/app/Console/Commands/DeployLocal.php
@@ -28,7 +28,12 @@ class DeployLocal extends Command
     public function handle(): int
     {
         $this->call('storage:link');
-
+        $this->call('optimize:clear');
+        $this->call('icons:clear');
+        $this->call('icons:cache');
+        $this->call('event:cache');
+        $this->call('optimize');
+        $this->call('livewire:discover')
         return 0;
     }
 }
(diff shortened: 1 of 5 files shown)


app/Console/Commands/DeployLocal.php
```diff
@@ -28,7 +28,12 @@ class DeployLocal extends Command
     public function handle(): int
     {
         $this->call('storage:link');
-
+        $this->call('optimize:clear');
+        $this->call('icons:clear');
+        $this->call('icons:cache');
+        $this->call('event:cache');
+        $this->call('optimize');
+        $this->call('livewire:discover')
```
jobara: Looks like you are missing a `;` here which may be what's causing the PR checks to fail.

```suggestion
        $this->call('livewire:discover');
```

jobara (approved):

PR #1615, opened 2023-03-28
test(route): added routes for status testing

### LINKED ISSUE #1602: SQLSTATE[HY000] [2002] Connection refused (SQL: select * from `sessions` where `id` = w67nFAnxed077E...
**`Illuminate\Database\QueryException`**

SQLSTATE[HY000] [2002] Connection refused (SQL: select * from `sessions` where `id` = w67nFAnxed077E2B750CHNiVwCdXJdO5xgGembsH limit 1/*controller='RedirectController',action='__invoke',url='%2F',route='generated%3A%3AlUGnUf3YYG6WkkwN'*/;)

:dart: Illuminate\Database\Connection::runQueryCallback
:page_facing_up: /app/vendor/laravel/framework/src/Illuminate/Database/Connection.php

```php
758  // lot more helpful to the developer instead of just the database's errors.
759  catch (Exception $e) {
760*     throw new QueryException(
761          $query, $this->prepareBindings($bindings), $e
762      );
```
[Open in Flare](https://flareapp.io/errors/3482386/latest)

marvinroman:
## Intention 
- Provide multiple status check options for pod status (meant only for internal Kube checks).
- Intend to only use `/status/db`. While we were at it we thought we could provide other potential future checks.

Created new status routes.

NGINX route `/nginx_status` will confirm that nginx is running.
Laravel routes
- `/status` checks that route was able to be reached checks NGINX & PHP.
- `/status/db` checks that route was able to be reached and database connection was made, checks NGINX, PHP & MySQL connection.

Added new route prefix `status` and 2 routes within it.
Added stub-status path in the NGINX configuration and limited to kube networking.

Branched from **staging** and merging back into **staging**.


app/Providers/RouteServiceProvider.php
```diff
@@ -46,6 +46,10 @@ public function boot()
             Route::middleware('web')
                 ->namespace($this->namespace)
                 ->group(base_path('routes/web.php'));
+
+            Route::prefix('status')
+                ->namespace($this->namespace)
+                ->group(base_path('routes/status.php'));
```
jobara, line 52: I'm not sure it's necessary to put this here. You can add prefixes directly in the routes file. Also by having these separate from web I'm not sure if they'll run through the wed middleware. Which may or may not be desirable. Could you explain a bit more about the implementation?

jobara:
@marvinroman I'm not sure what the best practice is for these, or all the intentions of the implementation you've submitted here. If you could explain a bit more about the intentions and functionality I think that would help. My current questions/comments are:

- Does this need to go through web routes? Can we do the checks using the API endpoints instead? In that way we can also set access permissions on those routes and not expose them through the website.
- Have you looked into @gtirloni's comment about using Kubernetes checks? (https://github.com/accessibility-exchange/platform/issues/1602#issuecomment-1486813267)

.deploy/accessibility-app/nginx.conf
```diff
@@ -50,15 +50,21 @@ http {
             access_log off;
             log_not_found off;
         }
-    
+
+        location /nginx_status {
+            stub_status on;
+            allow 10.0.0.0/8;
```
gtirloni: Could you add a comment saying where this range comes from (so we have a clue when things break)?

routes/status.php
```diff
@@ -0,0 +1,34 @@
...
+        return Response("Error connecting to database.", 500)
+        ->header('Content-Type', 'text/plain');
+    }
+});
+
+// plain web check NGINX & PHP working
+Route::get('/', function() {
+    return Response("Connection successful", 200)
+    ->header('Content-Type', 'text/plain');
+});
```
gtirloni, line 34: I think this will get confusing fast if issues happen. You need to remember that one healthcheck is only a no-op ping that only says the app not stuck and the other is actually a DB connection only (even this one won't alert if the database is gone, maybe it can be improved with an actual lightweight query that only retrieves a single record or only describes some important table, etc..).

Additionally, which one do you plan to use as the liveness check and which one as the readiness check? The DB or the no-op ping?

If we introduce Redis or some other component like an API that the app really depends on, are we adding other `/status/$path` endpoints?

app/Providers/RouteServiceProvider.php
marvinroman: I wanted them separate to minimize the impact of the  checks. These are just meant to be simple pod service checks internal to our Kube cluster to test whether routing should be done to the pod. 

I added intentions to the the original comment.

These aren't meant for testing full application health which will be done through external pings for full page rendering.

.deploy/accessibility-app/nginx.conf
marvinroman: I will push a comment, but it's the internal K8 network since it's only meant as a potential K8 healthcheck. Although we don't intend to use at this time.

routes/status.php
marvinroman: At this point we intend to use `/status/db` for both readiness and liveness probes.

The other routes are if we decide to separate the 2 in the future.

As far as increasing the depth of the database checks I don't see the need for an internal K8 check for the application pod. If we are concerned about database health, I would probably do those checks directly to the MySQL pod rather than going through the application pod.

We will also have external probes that will be checking full page rendering.

marvinroman:
@jobara In regards to going through API is that because you want authentication for status checks?
If you want external API healthchecks I would just do that in an external probe. I don't see it being necessary for the microservice health.

If you want these status checks restricted I would just make the following addition to the NGINX configuration.
```nginx
location /status {
    allow 10.0.0.0/8; # Internal K8 network
    deny all;
    try_files $uri $uri/ /index.php?$query_string;
}
```

jobara:
> @jobara In regards to going through API is that because you want authentication for status checks? If you want external API healthchecks I would just do that in an external probe. I don't see it being necessary for the microservice health.
> 
> If you want these status checks restricted I would just make the following addition to the NGINX configuration.
> 
> ```nginx
> location /status {
>     allow 10.0.0.0/8; # Internal K8 network
>     deny all;
>     try_files $uri $uri/ /index.php?$query_string;
> }
> ```

Really I just don't want to expose those endpoints over the web. That is I don't want someone to be able to hit accessibility exchange.ca/status/db. Although we may want to put up a status.accessibilityexchange.ca site separately that reports the health status.

marvinroman:
> > @jobara In regards to going through API is that because you want authentication for status checks? If you want external API healthchecks I would just do that in an external probe. I don't see it being necessary for the microservice health.
> > If you want these status checks restricted I would just make the following addition to the NGINX configuration.
> > ```nginx
> > location /status {
> >     allow 10.0.0.0/8; # Internal K8 network
> >     deny all;
> >     try_files $uri $uri/ /index.php?$query_string;
> > }
> > ```
> 
> Really I just don't want to expose those endpoints over the web. That is I don't want someone to be able to hit accessibility exchange.ca/status/db. Although we may want to put up a status.accessibilityexchange.ca site separately that reports the health status.

marvinroman:
> Really I just don't want to expose those endpoints over the web. That is I don't want someone to be able to hit accessibility exchange.ca/status/db. Although we may want to put up a status.accessibilityexchange.ca site separately that reports the health status.

We can very easily limit the status check to just he K8 network. I have included that limit in the changes to the NGINX config.

As far as status.accessibilityexchange.ca I did want to propose using Gitlab CI/CD to generate a status page hosted on Gitlab pages that would report statuses. That would make it separate from the cluster infrastructure so they can be reported if the cluster has some downtime. We could report statistics from Flare and external probes.

jobara:
> As far as status.accessibilityexchange.ca I did want to propose using Gitlab CI/CD to generate a status page hosted on Gitlab pages that would report statuses. That would make it separate from the cluster infrastructure so they can be reported if the cluster has some downtime. We could report statistics from Flare and external probes.

As long as we make sure not to expose anything that could lead to security issues this sounds good. Also, is it possible to do this in GitHub instead of GitLab so that we don't have as many services to manage?

marvinroman:
I will look into Github's offerings and process as far as that goes. I recommended Gitlab because of my own familiarity and our ops team already utilizes Gitlab for some of our CI/CD processes on our clusters already.

jobara:
@marvinroman just to let you know, I've changed this to merge to the Dev branch. I'm trying to have all the changes, if possible, go to dev first, and migrate to staging and production with deployments. In this way we make sure that the dev branch has all the changes that are in staging and production.

jobara:
@marvinroman could you please update this PR with the latest from the dev branch. Also, is there anything else we're waiting on for it. For example is there work needed for that independent status page or will that come in a separate PR?

marvinroman:
@jobara dev branch has been merged in.

app/Providers/RouteServiceProvider.php
jobara: Is it possible to not have these exposed through the site's URLs? right now I can navigate to the `/status/db` in a browser and it shows the status response.

marvinroman: Hi @jobara,

It won't work in your dev testing because the limits are in the webserver configuration which is different from the development environment you are using.

See `.deploy/accessibility-app/nginx.conf`  
```nginx
        # Status endpoints in Laravel
        location /status {
            allow 10.0.0.0/8; # Internal K8 network
            deny all;
            try_files $uri $uri/ /index.php?$query_string;
        }

        # Status endpoint for NGINX
        location /status/nginx {
            allow 10.0.0.0/8; # Internal K8 network
            deny all;
            stub_status on;
        }
```

jobara (approved):

PR #1725, opened 2023-05-30
1721 automate the push from GitHub to gitlab

### LINKED ISSUE #1721: Automate the push from github to gitlab
It would be ideal to include this as part of the Kube transition. The goal is to have mirroring repositories. See notes here for greater context: https://colabcoop.atlassian.net/wiki/spaces/IRIS/blog/2023/05/09/2615115777/IRIS+Full+Team+Sync+Release+Planning+May+2023

@JureUrsic - you communicated during the meeting that more research would be needed for this item. Can you also please estimate?

marvinroman:
## Description
This pull request adds a GitHub Actions workflows to mirror the `dev`, `staging` & `production` branch and tags to GitLab.

## Changes Made
- Created a new GitHub Actions workflow file: `.github/workflows/mirror-dev.yml`.
- Created a new GitHub Actions workflow file: `.github/workflows/mirror-staging.yml`.
- Created a new GitHub Actions workflow file: `.github/workflows/mirror-production.yml`.
- Added a workflow to mirror the  `dev`, `staging` & `production` branch and tags to GitLab.
- Restricted the workflow to run only on the `accessibility-exchange/platform` repository.

## Testing
- Tested the workflow on a fork to ensure it successfully mirrors the branches to GitLab without limit for repository.
- After adding limit for repository made sure it doesn't work on fork.
- Verified that tags are mirrored as well.

## Checklist
- [x] The workflow file is located in the correct directory: `.github/workflows/mirror-staging.yml`.
- [x] The workflow is correctly restricted to the `accessibility-exchange/platform` repository.
- [ ] Git global user name and email are correctly set.
- [ ] The GitLab remote URL is correctly configured.
- [ ] SSH key setup is working and allows pushing to GitLab.
- [ ] The workflow has been tested and verified for successful execution

### DIFF
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -27,5 +27,5 @@ jobs:
 
       - name: Push to GitLab
         run: |
-          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+          ssh-keyscan git.kube.v1.colab.coop >> ~/.ssh/known_hosts
           git push --atomic gitlab dev:development
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -1,0 +1,36 @@
+name: Mirror branches to GitLab
+
+on:
+  push:
+    branches:
+      - dev
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - dev
+  repository:
+    paths:
+      - 'accessibility-exchange/platform/**'
+
+jobs:
+  mirror:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Git
+        run: |
+          git config --global user.name "${{ vars.GIT_USER }}"
+          git config --global user.email "${{ vars.GIT_EMAIL }}"
+          git remote add gitlab ${{ vars.GITLAB_URL }}
+
+      - name: Push to GitLab
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+          git push gitlab HEAD:development
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   mirror:
-    if: github.repository == 'accessibility-exchange/platform'
+    # if: github.repository == 'accessibility-exchange/platform'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   mirror:
-    # if: github.repository == 'accessibility-exchange/platform'
+    if: github.repository == 'accessibility-exchange/platform'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -4,11 +4,6 @@ on:
   push:
     branches:
       - dev
-    tags:
-      - '*'
-  pull_request:
-    branches:
-      - dev
 
 jobs:
   mirror:
@@ -31,4 +26,4 @@ jobs:
           echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
           chmod 600 ~/.ssh/id_rsa
           ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
-          git push gitlab HEAD:development
+          git push --atomic gitlab dev:development
--- a/.github/workflows/mirror-dev.yml
+++ b/.github/workflows/mirror-dev.yml
@@ -20,10 +20,12 @@ jobs:
           git config --global user.email "${{ vars.GIT_EMAIL }}"
           git remote add gitlab ${{ vars.GITLAB_URL }}
 
+      - name: Setup SSH
+        uses: webfactory/ssh-agent@v0.7.0
+        with:
+            ssh-private-key: ${{ secrets.GITLAB_SSH_KEY }}
+
       - name: Push to GitLab
         run: |
-          mkdir -p ~/.ssh
-          echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
-          chmod 600 ~/.ssh/id_rsa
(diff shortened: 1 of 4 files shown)


marvinroman:
@jobara This is pending setup of Secrets and Variables.

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,34 @@
...
+          git config --global user.name "${{ vars.GIT_USER }}"
+          git config --global user.email "${{ vars.GIT_EMAIL }}"
+          git remote add gitlab ${{ vars.GITLAB_URL }}
+
+      - name: Push to GitLab
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
```
jobara: Instead of setting up the ssh key like this, I wonder if an approach as documented in this blog post ( https://www.webfactory.de/blog/use-ssh-key-for-private-repositories-in-github-actions ) would be better. Supposedly it doesn't write the private key to disk which will add a bit more security. They have also created a GitHub action you can use, if you'd like to investigate that.

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,34 @@
...
+
+on:
+  push:
+    branches:
+      - dev
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - dev
```
jobara: Could you please provide some info on why we'd need to run this when a PR is submitted?

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,34 @@
...
+          git config --global user.email "${{ vars.GIT_EMAIL }}"
+          git remote add gitlab ${{ vars.GITLAB_URL }}
+
+      - name: Push to GitLab
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+          git push gitlab HEAD:development
```
jobara: I'm not sure this is actually pushing tags. Following from this stack overflow post ( https://stackoverflow.com/a/3745250 ). I think what you'd like to do here is:

```bash
git push --atomic gitlab dev:development --follow-tags
```

jobara (changes requested):

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Git
+        run: |
+          git config --global user.name "${{ vars.GIT_USER }}"
+          git config --global user.email "${{ vars.GIT_EMAIL }}"
+          git remote add gitlab ${{ vars.GITLAB_URL }}
```
jobara, line 21: Are these vars supposed to be secrets?

.github/workflows/mirror-dev.yml
marvinroman: Looked further into PR trigger and removed.

.github/workflows/mirror-dev.yml
marvinroman: Are we sure we want to mirror tags? 

I don't see them included as triggers for deployment.

.github/workflows/mirror-dev.yml
marvinroman: Thanks this was a great reference. I will be changing to use their action.

.github/workflows/mirror-dev.yml
marvinroman: No. I intended that they would be variables and not secrets since they don't need to be masked for security.

This makes it easier to review the action logs to determine the values being set.

.github/workflows/mirror-dev.yml
marvinroman: I included `.github/workflows/mirror-tags.yml` for a look a way to mirror tags.

.github/workflows/mirror-dev.yml
jobara: Where do the vars get sources from?

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Push to GitLab
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.GITLAB_SSH_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
```
jobara: You've used `webfactory/ssh-agent@v0.7.0` in the other workflows, did you mean to use that here too?

.github/workflows/mirror-dev.yml
jobara: > Are we sure we want to mirror tags?
> 
> I don't see them included as triggers for deployment.

It just depends if you want a copy of the tags in GitLab or not.

.github/workflows/mirror-dev.yml
marvinroman: Same area as secrets but different tab.
Screenshot_20230531_181754

.github/workflows/mirror-dev.yml
marvinroman: Thanks for catching that. I should have reviewed after merging stashed changes.

.github/workflows/mirror-dev.yml
marvinroman: I can leave the workflow. It just has no impact as far as GitLab CI/CD automations at this point.

.github/workflows/mirror-dev.yml
```diff
@@ -0,0 +1,31 @@
...
+          git remote add gitlab ${{ vars.GITLAB_URL }}
+
+      - name: Setup SSH
+        uses: webfactory/ssh-agent@v0.7.0
+        with:
+            ssh-private-key: ${{ secrets.GITLAB_SSH_KEY }}
+
+      - name: Push to GitLab
+        run: |
+          ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
```
jobara: Is `gitlab.com` correct here, given that we are using a self hosted instance?

.github/workflows/mirror-dev.yml
jobara: I didn't see these in the repo so I've added them. However, please update them as needed.

.github/workflows/mirror-dev.yml
marvinroman: Updated the URL.

jobara (approved):

PR #1869, opened 2023-07-24
ci: terraform is creating commits on gitlab causing push issues (resolves #1866)

### LINKED ISSUE #1866: Terraform is creating commits on Gitlab causing push issues
When there is a version update, Terraform will push a commit to update versions. This new commit in Gitlab puts Github behind on commits and prevents future commits from automatically being pushed.

marvinroman:
Adds a pull request to pull any changes on Gitlab for the branch before attempting to push.

* [x] testing on fork and testing repo on gitlab to ensure that changes were merged in before push.


jobara:
@marvinroman the thing I'm concerned with this approach is that we never have all the changes in the GitHub repo. There will be changes that are only ever in the GitLab repo. This means that local development and our canonical repository will not have the same codebase as the deployed instance. Although the changes in this case (such as https://github.com/accessibility-exchange/platform/pull/1869/commits/3f8e53048e0f016e514c71ed53c8924cd19880b0 ) wouldn't be things we use in local development, it would be confusing when we investigate issues and when others are looking at the code through GitHub. 

Is it possible for this bit of configuration to be pulled in from an Environment variable instead and for terraform to update that. Or some other method that wouldn't require commits to the repo. At least in that way we wouldn't have this commit issue and would have an expectation that it is part of the deploy environment.

Are there other options that we can explore that would remove the commit issue and allow for tracking of the docker image?

marvinroman:
@JureUrsic can you answer and review Justin's suggestion above?

PR #1926, opened 2023-10-10
fix: path segment names are not all localizable

### LINKED ISSUE #1922: localize all path segment names
**Prerequisites**

**Describe the bug**

Not all of the routes have localized names. They should be defined and lang/[lang]/route.php file. For example https://github.com/accessibility-exchange/platform/blob/dev/resources/lang/fr/routes.php

**To Reproduce**

For example:

https://dev.accessibilityexchange.ca/en/about/for-individuals and https://dev.accessibilityexchange.ca/fr/about/for-individuals

vs

https://dev.accessibilityexchange.ca/en/register and https://dev.accessibilityexchange.ca/fr/inscription

**Expected behavior**

Route paths should be localized.

jobara:
### Prerequisites

- [ ] I've run `composer analyze` and addressed any static analysis issues.
- [ ] I've run `php artisan test` and ensured that all tests pass.


greatislander (changes requested):
Routes can't be translated in to ASL or LSQ. I note there is no ASL file. Should it just be duplicated from English?

jobara:
> Routes can't be translated in to ASL or LSQ. I note there is no ASL file. Should it just be duplicated from English?

I'm not 100% sure I'm understanding this correctly, so please let me know if I've misunderstood. I believe if there is no localization provided that it will fall back to the `key` which is in English so the ASL related routes should be the same as EN. Regarding LSQ this should be copied from FR when the localize command runs the [folder copy action](https://github.com/accessibility-exchange/platform/blob/c0f88f4fb53787b236b3e3bdeec2062e5e976f63/composer.json#L121). This will use the French localization for the Routes to match the French text used on the pages when LSQ is set.

greatislander:
Ah, you're right. Thanks for the reminder of how that works!

greatislander (approved):

PR #1935, opened 2023-10-13
feat: automate resetting dev deploy (resolves #1902, #1903)

### LINKED ISSUE #1902: Automate resetting Dev deploy

The original Dev deploy would reset nightly; however, this feature wasn't included during the migration to the CoLab infrastructure. This means that the dev site retains all of the various changes as it is being tested.

The Dev deploy should have the database reset nightly. That way each day starts with a fresh start with the dev seeder run.

**Describe alternatives you've considered**

Leaving this as a manual process.

**Additional context**

Locally I'd run `php artisan migrate:fresh --seeder=DevSeeder` to reset the state. Something similar could be run as part of an automated process. Possibly using Laravel's [Task Scheduling](https://laravel.com/docs/9.x/scheduling).

We'd also like to reset the Staging deploy, but this will be when deploying release candidates. Not sure if this can or should be automated. For now it could be done with a manual process when making the release candidates. If there is an automated solution, we should file an issue for that as well.

### LINKED ISSUE #1903: Provide a mechanism for updating staging and dev with data from production

It would be nice to be able to test changes with real data before pushing to production. Being able to pull data from production into staging and dev deploys would help to get a better idea of how new features and fixes affect real data.

Be able to automatically migrate or have an ability to fetch certain database values from production to populate into staging and dev. This would only be for data that is entered by the service org such as resources, topics, site settings, sign language interpretations and etc. It must **NOT** include user data and account information. 

**Describe alternatives you've considered**

Continue only using fake/sample data for use on the dev and staging site.

marvinroman:
Creates a scheduled reset 15 minutes after midnight every day for the **dev** environment.


app/Console/Kernel.php
```diff
@@ -28,6 +28,11 @@ protected function schedule(Schedule $schedule)
             ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
             ->onOneServer(); // run only on a single server at once
 
+        $schedule->command('migrate:fresh --seeder=DevSeeder') // use custom command to make sure that the commands are chained
+            ->dailyAt('00:15') // Run daily at 12:15 am
+            ->environments(['dev']) // only run for APP_ENV tagged staging, dev, or local
```
jobara: It seems the comment needs to be updated.

app/Console/Kernel.php
```diff
@@ -28,6 +28,11 @@ protected function schedule(Schedule $schedule)
             ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
             ->onOneServer(); // run only on a single server at once
 
+        $schedule->command('migrate:fresh --seeder=DevSeeder') // use custom command to make sure that the commands are chained
+            ->dailyAt('00:15') // Run daily at 12:15 am
```
jobara, line 33: Do you know which timezone this is?

marvinroman: What timezone would you like and I will declare it explicitly. I am sure it probably uses server time by default.

app/Console/Kernel.php
marvinroman: Updated.

app/Console/Kernel.php
jobara: Could you set it to PT so that its update around midnight pacific time.

marvinroman: Done.

app/Console/Kernel.php
```diff
@@ -26,10 +26,18 @@ protected function schedule(Schedule $schedule)
         $schedule->command('db:seed:backup --all') // use custom command to make sure that the commands are chained
             ->daily() // Run daily at midnight
             ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
+            ->timezone('America/Los_Angeles') // Run as PST timezone
+            ->onOneServer(); // run only on a single server at once
+
+        $schedule->command('migrate:fresh --seeder=DevSeeder') // use custom command to make sure that the commands are chained
```
jobara: Is it possible to put the site into maintenance mode while this is happening?

marvinroman: Created new `app:refresh-dev` command that allows us to chain maintenance mode into the refresh.
Also migrated #1903 into this and closed out the associated PR.

marvinroman:
@jobara for the failing tests I will reach out to you tomorrow with the variables that we need to add into Github for the tests to pass.

app/Console/Commands/RefreshDev.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        $this->call('down');
+        if (in_array(config('app.env'), ['dev']) !== true) {
+            $this->call('migrate:fresh --seeder=DevSeeder');
+        }
```
jobara: I'm not sure I understand this step. It looks like it will clear the database only when it isn't in the dev environment. Which I think would be the opposite of what we'd want. Am I misunderstanding this?

app/Console/Kernel.php
```diff
@@ -25,11 +25,19 @@ protected function schedule(Schedule $schedule)
...
             ->daily() // Run daily at midnight
+            ->environments(['staging', 'dev', 'local', 'production']) // only run for APP_ENV tagged staging, dev, local, or production
+            ->timezone('America/Los_Angeles') // Run as PST timezone
+            ->onOneServer(); // run only on a single server at once
+
+        $schedule->command('app:refresh-dev') // use custom hcommand to make sure that te commands are chained
+            ->dailyAt('00:15') // Run daily at 12:15 am
             ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
+            ->timezone('America/Los_Angeles') // Run as PST timezone
             ->onOneServer(); // run only on a single server at once
```
jobara: I guess this gets added to all deployments. Is it possible to only add this to the scheduling on the dev deploy?

app/Console/Commands/RefreshDev.php
marvinroman: That was an old restriction on the scheduler. It only ran on the **dev** environment so I kept that restriction in the command steps.
I we want to to remove restriction we can and we can invert the logic to just make sure it does not run on production?

app/Console/Kernel.php
marvinroman: Yeah we can.

app/Console/Commands/RefreshDev.php
jobara: But doesn't the code say run everywhere but 'dev'?

jobara:
> @jobara for the failing tests I will reach out to you tomorrow with the variables that we need to add into Github for the tests to pass.

As discussed its preferable to not require the ENVs for use on GitHub and when running locally.

app/Console/Kernel.php
marvinroman: Limited to **dev** now.

app/Console/Commands/RefreshDev.php
marvinroman: I changed this to limit it from running on **testing** and **production**.

```php
        if (in_array(config('app.env'), ['testing', 'production']) !== true) {
            $this->call('down');
            $this->call('migrate:fresh --seeder=DevSeeder');
            $this->call('db:seed:backup --all --restore --from=production');
            $this->call('up');
        }
```

app/Console/Commands/SeederBackupData.php
```diff
@@ -38,14 +38,23 @@ public function handle()
         $options = $this->options();
 
         // option to use environment to restore or backup to different environment files
-        if (isset($options['env']) && in_array($options['env'], config('backup.filament_seeders.environments')) === true) {
```
marvinroman, line 41: This is to resolve the passing of a separate seeders environment on to the seeders to make sure it resolves the filename correctly.

app/Console/Kernel.php
```diff
@@ -25,11 +25,19 @@ protected function schedule(Schedule $schedule)
     {
         $schedule->command('db:seed:backup --all') // use custom command to make sure that the commands are chained
             ->daily() // Run daily at midnight
-            ->environments(['staging', 'dev', 'local']) // only run for APP_ENV tagged staging, dev, or local
+            ->environments(['staging', 'dev', 'local', 'production']) // only run for APP_ENV tagged staging, dev, local, or production
+            ->timezone('America/Los_Angeles') // Run as PST timezone
+            ->onOneServer(); // run only on a single server at once
+
+        $schedule->command('app:refresh-dev') // use custom hcommand to make sure that te commands are chained
```
jobara: typo "hcommand"

database/seeders/IdentitySeeder.php
```diff
@@ -18,15 +18,24 @@ public function run()
...
 
-        if (Storage::disk('seeds')->exists(sprintf('identities.%s.json', $environment))) {
+        // fix for when it runs in environments without access to S3 bucket
+        try {
+            // try connecting to the seeds S3 bucket
+            Storage::disk('seeds');
+        } catch (\Exception $e) {
+            // mock the seeds filesystem locally
+            Storage::fake('seeds');
+        }
```
jobara: As an alternative to this, why not have a local and s3 version of the seed disk (e.g. seed-local, seed-s3) and change between them based on an environment variable. For example have an ENV like `SEED_DRIVER` and in the example env set that to `seed-local` but in the deployment configuration set that to `seed-s3`. In that way you won't have to do all the exception checking in each seeder and you can run the backups locally; for example to test the backup system.

See: https://laracasts.com/discuss/channels/laravel/how-to-use-local-filesystem-for-development-but-s3-for-production

database/seeders/ResourceSeeder.php
```diff
@@ -29,15 +29,15 @@ public function run()
         }
 
         // TODO need to write handling of attachments
-        if (false && Storage::disk('seeds')->exists(sprintf('resources.%s.json', $environment))) {
-            // if truncate was set via seeder restore command then truncate the table prior to seeding data
+        if (false && Storage::disk('seeds')->exists(sprintf('%s/resources.%s.json', config('filesystems.disks.seeds.path'), $environment))) {
+            // if trucate was set via seeder restore command then truncate the table prior to seeding data
```
jobara: typo "trucate"

jobara: There seems to be other instances of this typo as well.

jobara (changes requested):

app/Console/Commands/SeederBackupData.php
```diff
@@ -38,14 +38,24 @@ public function handle()
...
+        // fix for when it runs in environments without access to S3 bucket
+        // TODO remove after testing
+        // try {
+        //     // try connecting to the seeds S3 bucket
+        //     Storage::disk('seeds');
+        // } catch (\Exception $e) {
+        //     // mock the seeds filesystem locally
+        //     Storage::fake('seeds');
+        // }
+
```
jobara: Can this be removed?

app/Console/Kernel.php
jobara: @marvinroman this typo is still present.

jobara (changes requested):

jobara (approved):

jobara:
@marvinroman because the remaining changes were minor, I made them myself and merged in.

PR #1937, opened 2023-10-17
fix: block pending users from seeing approved user only content

### LINKED ISSUE #1803: Unapproved accounts can view resources, projects and individual user's public pages
**Prerequisites**

**Describe the bug**

Only approved accounts should be able to view projects, public pages, and organization pages.

**To Reproduce**
1. Create a new account
2. Before being approved go to the "Projects" and "People and Organizations" tabs. Notice that you can view all of the pages even though the account hasn't been approved yet.

**Expected behavior**

The "Projects" and "People and Organization" links should not be present in the navigation and their entire route paths should not be viewable by unapproved accounts; probably returning 403 errors.

chosww:
Resolves the issue described in #1803 

viewAny rules used in many middleware were using rules specified in canViewPublishedContent method defined in UserCanViewPublishedContent, and the method was not guarding the case when users are not approved. Fix to the issue was by adding additional check to the method to see if specific user is oriented.

### Prerequisites


jobara:
@chosww I'm finding that when I create new FRO and Community org users that they can't see the projects and browse people even after being approved.

This may be related to the organizations being approved vs the users in these cases.

chosww:
> @chosww I'm finding that when I create new FRO and Community org users that they can't see the projects and browse people even after being approved.
> 
> This may be related to the organizations being approved vs the users in these cases.

Oh, thanks for catching that, I will investigate the issue.

chosww:
I spent some time to think about the work flow around organization user getting approved by the admin; it feels weird to me that when the user gets approved, it's the organization that gets approved but not the user itself. If we want to keep it that way, it makes sense to loop through all the users in the approved organization and update their oriented_at time, just like how we do that for suspend. One more thing I am wondering is what happens when new users join the organization? Do we check oriented_at status of organization and just set oriented_at of the new member on sign up? Would be good to know about what both of you think @jobara and @greatislander.

chosww:
Updating canViewPublishedContent this way make it work in our favour: 

```php
public function canViewPublishedContent(User $user): bool
    {
        $isOriented = $user->organization?->checkStatus('approved') || $user->regulatedOrganization?->checkStatus('approved') || $user->oriented_at != null;
        return
            $user->hasVerifiedEmail() &&
            $isOriented &&
            in_array(
                $user->context,
                [
                    UserContext::Administrator->value,
                    UserContext::Individual->value,
                    UserContext::Organization->value,
                    UserContext::RegulatedOrganization->value,
                ]
            );
    }

```

greatislander:
This looks good to me but I will defer to @jobara.

jobara:
@chosww do you still have more tests that you wanted to add to this?

chosww:
> @chosww do you still have more tests that you wanted to add to this?

I am trying to write tests for when organization or regulated organizations are oriented, their users can view the pages, but I am having trouble making test working :/

tests/Feature/IndividualTest.php
```diff
@@ -806,27 +769,51 @@
...
-    $response->assertForbidden();
+    actingAs($pendingUser)->get(localized_route('individuals.index'))->assertForbidden();
 
     $pendingUser->update(['oriented_at' => now()]);
-    $response = $this->actingAs($pendingUser)->get(localized_route('individuals.index'));
-    $response->assertOk();
+    actingAs($pendingUser)->get(localized_route('individuals.index'))->assertOk();
+});
+
+test('organization or regulated organization users can not view individual pages if they are not oriented', function () {
```
chosww, line 791: @jobara lines 791 to 816 are new tests I added.

tests/Feature/OrganizationTest.php
```diff
@@ -1067,23 +1009,47 @@
...
-    $response->assertForbidden();
+    actingAs($pendingUser)->get(localized_route('organizations.index'))->assertForbidden();
 
     $pendingUser->update(['oriented_at' => now()]);
-    $response = $this->actingAs($pendingUser)->get(localized_route('organizations.index'));
-    $response->assertOk();
+    actingAs($pendingUser)->get(localized_route('organizations.index'))->assertOk();
+});
+
+test('organization or regulated organization users can not view organizations if they are not oriented', function () {
```
chosww, line 1018: @jobara lines 1018 to 1043 are new tests I added.

tests/Feature/ProjectTest.php
```diff
@@ -240,28 +225,65 @@
...
     $pendingUser->update(['oriented_at' => now()]);
 
-    $response = $this->actingAs($pendingUser)->get(localized_route('projects.my-projects'));
-    $response->assertOk();
+    actingAs($pendingUser)->get(localized_route('projects.my-projects'))->assertOk();
+
+    actingAs($pendingUser)->get(localized_route('projects.all-projects'))->assertOk();
+});
+
+test('organization or regulated organization users can not view projects if they are not oriented', function () {
```
chosww: @jobara lines 245 to 286 are new tests I added.

tests/Feature/RegulatedOrganizationTest.php
```diff
@@ -654,38 +635,59 @@
...
-    $response->assertForbidden();
+    actingAs($pendingUser)->get(localized_route('regulated-organizations.index'))->assertForbidden();
 
     $pendingUser->update(['oriented_at' => now()]);
-    $response = $this->actingAs($pendingUser)->get(localized_route('regulated-organizations.index'));
-    $response->assertOk();
+    actingAs($pendingUser)->get(localized_route('regulated-organizations.index'))->assertOk();
+});
+
+test('organization or regulated organization users can not view regulated organizations if they are not oriented', function () {
```
chosww, line 654: @jobara lines 654 to 679 are new tests I added.

tests/Feature/IndividualTest.php
```diff
@@ -408,12 +392,11 @@
...
     expect($individual->user->support_person_vrs)->toBeNull();
 
-    $response->assertSessionHasNoErrors();
-    $response->assertRedirect(localized_route('individuals.edit', ['individual' => $individual, 'step' => 4]));
+    $response->assertSessionHasNoErrors()->assertRedirect(localized_route('individuals.edit', ['individual' => $individual, 'step' => 4]));
 });
 
 test('entity users can not create individual pages', function () {
-    $user = User::factory()->create(['context' => 'regulated-organization']);
+    $user = User::factory()->create(['context' => 'regulated- g ']);
```
jobara: Was this a mistake?

jobara (changes requested):
@chosww so one thing is that as an FRO or Community Org I can start creating a project before I'm approved; which is expected. However, unless I manually keep track of the URL, I won't be able to get back to it as I can't access my own project listings. Perhaps we should still allow for accessing the projects page, but not the one to browse all and other orgs projects. What do you think? Also wonder if @greatislander and @cherylhjli have any thoughts on this.

resources/views/projects/my-projects.blade.php
```diff
@@ -79,18 +81,26 @@
...
-    @endswitch
+            @default
+                @include(isset($section) ? 'projects.my-projects.' . $section : 'projects.my-projects.participating')
+        @endswitch
+    @else
+        @if ($user->context === 'regulated-organization')
+            @includeWhen($projectable, 'projects.my-projects.running')
+        @elseif ($user->context === 'organization')
+            @include('projects.my-projects.running')
+        @endif
```
jobara: If I'm reading this correctly, for regulated-organizations you only include the `project.my-projects.running` partial when there is a `$projectable` but for organizations you always include it. Is that correct and if so why the distinction?

I also wonder if you could collapse this down into a single `@includeWhen` statement.

jobara:
The `browse all projects` section and links should not be shown when the organization/fro does not have access to view all projects.

Screenshot showing browse all links when you don't have access to that page

resources/views/projects/my-projects.blade.php
chosww: Thanks @jobara for pointing this out, I just followed what was there before, and only regulated-organization was using includeWhen to show the running blade. @includeWhen seems to be working for organization account as well, so I will update this to use single @includeWhen.

jobara:
@chosww there's still a "Browse all projects" link towards the top right. It's to the right of the heading.
Screenshot showing the browse all projects link

jobara (approved):

PR #1957, opened 2023-11-01
feat: allow ToS and Privacy Policy to be edited

jobara:
- Migrates ToS and Privacy Policy to be editable in a Filament admin form.
- Restores prefixes to route groups that were removed in #1926 

### Prerequisites

If this PR changes PHP code or dependencies:

- [x] I've run `composer format` and fixed any code formatting issues.
- [x] I've run `composer analyze` and addressed any static analysis issues.
- [x] I've run `php artisan test` and ensured that all tests pass.
- [x] I've run `composer localize` to update localization source files and committed any changes.

If this PR changes CSS or JavaScript code or dependencies:

- [ ] I've run `npm run lint` and fixed any linting issues.
- [ ] I've run `npm run build` and ensured that CSS and JavaScript assets can be compiled.

PR #1959, opened 2023-11-07
feat: add sitemap and robots (resolves #1388)

### LINKED ISSUE #1388: Update Robots.txt to allow proper site indexing for live site

chosww:
### Prerequisites


app/Console/Commands/GenerateRobots.php
```diff
@@ -0,0 +1,32 @@
...
+use Illuminate\Console\Command;
+
+class GenerateRobots extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'app:generate-robots';
```
jobara: Is there a reason for calling this `app:generate-robots`? I noticed for the sitemap you called it `sitemap:generate`. I wonder if they can share a similar convention and possibly be in the same namespace.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $routes = ['/' => ['en' => 'en', 'asl' => 'asl', 'fr' => 'fr', 'lsq' => 'lsq']];
+        // once deployed to the server, files have the same modified date, use README as a default last modified date
+        $lastMod = Carbon::createFromTimestamp(filemtime('./README.md'))->toISOString();
```
jobara: It would likely be a bit clearer throughout if this had the same name as the variable used in the template.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,50 @@
...
+        foreach (Route::getRoutes()->get('GET') as $route) {
+            $routeURI = $route->uri();
+            if (str_contains($routeURI, 'about')) {
```
jobara: Probably should use starts with if possible instead of contains, as we don't want false positives. You could look at [Str::startsWith()](https://laravel.com/docs/8.x/helpers#method-starts-with).

app/Console/Commands/GenerateRobots.php
```diff
@@ -0,0 +1,32 @@
...
+     * @var string
+     */
+    protected $description = 'Command description';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        file_put_contents('./public/robots.txt', view('robots')->render());
```
jobara, line 28: Should add this path to the gitignore.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,50 @@
...
+                [$locale, $url] = explode('/', $routeURI, 2);
+                if (array_key_exists($url, $routes)) {
+                    $routes[$url][$locale] = $routeURI;
+                } else {
+                    $routes[$url] = [$locale => $routeURI];
+                }
+            }
+        }
+        file_put_contents('./public/sitemap.xml', view('sitemap', ['routes' => $routes, 'default_lastmod' => $lastMod])->render());
```
jobara: Should add this path to the gitignore.

app/Console/Commands/GenerateRobots.php
```diff
@@ -0,0 +1,32 @@
...
+     * @var string
+     */
+    protected $signature = 'app:generate-robots';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Command description';
```
jobara: Should add a real description.

resources/views/robots.blade.php
```diff
@@ -0,0 +1,8 @@
+<?php header('Content-Type: text/plain; charset=utf-8'); ?>
+Sitemap: {{ env('APP_URL') }}/sitemap.xml
+
+User-Agent: *
+Disallow: /
```
jobara: Did you mean to disallow crawling the whole site? I'm wondering because below there is a specific case to disallow GPTBot.

resources/views/sitemap.blade.php
```diff
@@ -0,0 +1,19 @@
+<?= '<' . '?' . 'xml version="1.0" encoding="UTF-8"?>' . "\n" ?>
```
jobara, line 1: Probably related to how the content in the template was rendered but I'm not exactly sure because the output seems to have even more indents. It's likely not really a big deal but just noting that the output is hard to read.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml"
    xmlns:image="http://www.google.com/schemas/sitemap-image/1.1"
    xmlns:video="http://www.google.com/schemas/sitemap-video/1.1"
    xmlns:news="http://www.google.com/schemas/sitemap-news/0.9">
            <url>
                                                <loc>http://platform.test/en</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/terms-of-service</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/terms-of-service"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/terms-of-service"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/terms-of-service"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/privacy-policy</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/privacy-policy"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/privacy-policy"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/privacy-policy"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/consultation-participants</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/consultation-participants"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/consultation-participants"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/consultation-participants"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/consultation-participants/what-we-ask-for</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/consultation-participants/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/consultation-participants/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/consultation-participants/what-we-ask-for"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/accessibility-consultants</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/accessibility-consultants"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/accessibility-consultants"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/accessibility-consultants"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/accessibility-consultants/what-we-ask-for</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/accessibility-consultants/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/accessibility-consultants/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/accessibility-consultants/what-we-ask-for"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/community-connectors</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/community-connectors"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/community-connectors"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/community-connectors"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-individuals/community-connectors/what-we-ask-for</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-individuals/community-connectors/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-individuals/community-connectors/what-we-ask-for"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-individuals/community-connectors/what-we-ask-for"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-community-organizations</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-community-organizations"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-community-organizations"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-community-organizations"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-community-organizations/consultation-participants</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-community-organizations/consultation-participants"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-community-organizations/consultation-participants"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-community-organizations/consultation-participants"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-community-organizations/accessibility-consultants</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-community-organizations/accessibility-consultants"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-community-organizations/accessibility-consultants"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-community-organizations/accessibility-consultants"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-community-organizations/community-connectors</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-community-organizations/community-connectors"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-community-organizations/community-connectors"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-community-organizations/community-connectors"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-community-organizations/get-input</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-community-organizations/get-input"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-community-organizations/get-input"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-community-organizations/get-input"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-regulated-organizations</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-regulated-organizations"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-regulated-organizations"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-regulated-organizations"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/for-regulated-organizations/get-input</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/for-regulated-organizations/get-input"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/for-regulated-organizations/get-input"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/for-regulated-organizations/get-input"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/pricing</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/pricing"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/pricing"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/pricing"
                        rel="alternate" />
                                    </url>
            <url>
                                                <loc>http://platform.test/en/about/glossary</loc>
                    <lastMod>2023-11-02T12:14:42.000000Z</lastMod>
                                                                <link hreflang="asl" href="http://platform.test/asl/about/glossary"
                        rel="alternate" />
                                                                <link hreflang="fr" href="http://platform.test/fr/about/glossary"
                        rel="alternate" />
                                                                <link hreflang="lsq" href="http://platform.test/lsq/about/glossary"
                        rel="alternate" />
                                    </url>
    </urlset>
```

app/Console/Commands/GenerateSitemap.php
jobara: Actually I don't think this will work because eventually those routes are going to be localized. At the very least you'll need to look for `about` and the French translation of it. Which we haven't received yet. It does look like the routes are returned with them all grouped together though. There may also be other ways to find them.

jobara: Instead of looking at the route URI I think you can check the route name instead. This is the name we give to the route with `->name()`. So you can get `$route->getName()` and check that value instead as it should be consistent across locales with difference being the local prepended. so something named `about.glossary` would be returned as `en.about.glossary` and so forth if it had be created with `Route::multilingual`

See: https://laravel.com/api/10.x/Illuminate/Routing/Route.html#method_getName

jobara (changes requested):

app/Console/Commands/GenerateRobots.php
chosww: I forgot to update default values from "php artisan make:command"; I will update them with appropriate values.

app/Console/Commands/GenerateSitemap.php
chosww: That makes sense to use route name instead of uri. One thing I noticed is that glossary is not defined in web.php, so  it is not included when I use getName method. Should we add it as a special case like the main page?

resources/views/robots.blade.php
chosww: oh, I would like to get some suggestions around which pages we should disallow from the user-agents; I am thinking disallow every pages those are not /about/, but for now I left it as what it was before.

resources/views/sitemap.blade.php
chosww: Apparently, it includes all the indentations in the sitemap blade template. I removed some indentations to make the output properly formatted, but seems like auto formatting ran when I committed the code. Is it prettier that fixes the indentation? If so, could we include sitemap template to prettierignore?

app/Console/Commands/GenerateSitemap.php
jobara: There's a require call to bring it in through the web.php at the bottom of the file. I'm surprised it isn't showing up. Shouldn't that call be bringing in all the available routes? We should probably look into this a bit as we may have other things that are defined in separate files in the future.

resources/views/robots.blade.php
jobara: Probably anything that would give a 400 error should be blocked from crawling but the rest are okay. Actually I don't think it could reach those pages that are behind the login anyways. I guess explicitly disallow the following:

- admin/
- api/
- status/

Perhaps @greatislander would have some thoughts on this.

resources/views/sitemap.blade.php
jobara: Maybe it isn't worth the trouble to change. I guess just leave it for now if it works. Probably won't be reading it much and it isn't too hard to fix the formatting if we need to review it.

app/Console/Commands/GenerateSitemap.php
chosww: It seemed like we didn't include "about." in glossary route name. I added it and I can include glossary properly.

app/Console/Commands/GenerateRobots.php
```diff
@@ -0,0 +1,32 @@
...
+     * @var string
+     */
+    protected $signature = 'robots:generate';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Generate the robots.';
```
jobara: Given that robots could have different meanings, it might make sense to say robots.txt file

```suggestion
    protected $description = 'Generate the robots.txt file.';
```

app/Console/Kernel.php
```diff
@@ -31,6 +31,18 @@ protected function schedule(Schedule $schedule)
...
+            ->daily() // Run daily at midnight
+            ->environments(['staging', 'dev', 'local', 'production']) // only run for APP_ENV tagged staging, dev, local, or production
+            ->timezone('America/Los_Angeles') // Run as PST timezone
+            ->onOneServer(); // run only on a single server at once
+
+        $schedule->command('app:generate-robots') // generate robots
+            ->daily() // Run daily at midnight
+            ->environments(['staging', 'dev', 'local', 'production']) // only run for APP_ENV tagged staging, dev, local, or production
+            ->timezone('America/Los_Angeles') // Run as PST timezone
+            ->onOneServer(); // run only on a single server at once
```
jobara: Actually I think they need to be run on both servers because the filesystems aren't shared.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $routes = ['/' => ['en' => 'en', 'asl' => 'asl', 'fr' => 'fr', 'lsq' => 'lsq']];
+        // once deployed to the server, files have the same modified date, use README as a default last modified date
+        $default_lastmod = Carbon::createFromTimestamp(filemtime('./README.md'))->toISOString();
+        foreach (Route::getRoutes()->get('GET') as $route) {
+            if (str_contains($route->getName(), 'about')) {
```
jobara: I still think using `str_contains` is susceptible to false positives. How about using [`$route->named()`](https://laravel.com/api/10.x/Illuminate/Routing/Route.html#method_named). This takes in patterns that can be matched and returns a boolean value. So something like `$route->named('*.about.*')`. I think it might be better though to define this in a config file as an array, then you'd call something like `$route->named(config('sitemap.patterns'))` to test against the routes. You could even be more explicit and use the specific locales instead of the first wildcard if you'd like.

jobara (changes requested):
@chosww what do you think about changing the command signatures so they are grouped together? Something like `seo:generate-robots` and `seo:generate-sitemap`.

app/Console/Commands/GenerateSitemap.php
chosww: Thanks Justin, I forgot to update this part, I agree with you about str_contains can be false positive. I added a new config file called seo.php and took the approach you suggested.

jobara:
Now that #1957 has been merged the ToS and privacy policy are now model based and you should get the modified date from the model. I think you'll probably want to add a config for files to ignore when you do your initial loop through files to add to the sitemap. You'll want to ignore `about.terms-of-service`, `about.privacy-policy` and `about.page`. You'll then want to add these back into your routes array by looping through all `Page` models and getting their urls. You could probably call `localized_route()` on each of them for each locale. for example: `localized_route('about.pages', $page, 'en')` and the modified date would be from `$page->updated_at`. You may need to do the ToS and Privacy Policy pages separate from the other possible pages, as the URI is a bit different.

chosww:
> Now that #1957 has been merged the ToS and privacy policy are now model based and you should get the modified date from the model. I think you'll probably want to add a config for files to ignore when you do your initial loop through files to add to the sitemap. You'll want to ignore `about.terms-of-service`, `about.privacy-policy` and `about.page`. You'll then want to add these back into your routes array by looping through all `Page` models and getting their urls. You could probably call `localized_route()` on each of them for each locale. for example: `localized_route('about.pages', $page, 'en')` and the modified date would be from `$page->updated_at`. You may need to do the ToS and Privacy Policy pages separate from the other possible pages, as the URI is a bit different.

Thanks for the suggestion @jobara. I initially went with your suggestion of looping through all `Page` models, but then realized that could be omitted if we could query exact Page model we are interested in. Also, I wasn't too happy about how redundant the code blocks become, so I took a slight different approach. Please let me know if there's any issue you see from my approach.

resources/views/sitemap.blade.php
```diff
@@ -0,0 +1,21 @@
...
+    xmlns:image="http://www.google.com/schemas/sitemap-image/1.1"
+    xmlns:video="http://www.google.com/schemas/sitemap-video/1.1"
+    xmlns:news="http://www.google.com/schemas/sitemap-news/0.9">
+    @foreach ($routes as $baseUrl => $localizedUrls)
+        <url>
+            @if ($baseUrl !== 'about/page/{page}')
+                @foreach ($localizedUrls as $locale => $localizedUrl)
+                    @if ($locale === 'en')
+                        <loc>{{ env('APP_URL') . '/' . $localizedUrl }}</loc>
+                        <lastMod>{{ isset($lastmod[$baseUrl]) ? $lastmod[$baseUrl] : $lastmod['default'] }}</lastMod>
```
jobara: can this be written like: 

```suggestion
                        <lastMod>{{ $lastmod[$baseUrl] ?? $lastmod['default'] }}</lastMod>
```

resources/views/sitemap.blade.php
```diff
@@ -0,0 +1,21 @@
...
+                    @else
+                        <link hreflang="{{ $locale }}" href="{{ env('APP_URL') . '/' . $localizedUrl }}"
+                            rel="alternate" />
```
jobara: I'm not sure if it matters or what's actually better, but something to maybe think about would be to return the full URL instead of the relative URL in `$localizedUrls`. For example when generating you could do the prepending of `env('APP_URL')` or made a call to `localized_route()`.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,56 @@
...
+                $routeURI = $route->uri();
+                [$locale, $url] = explode('/', $routeURI, 2);
+                if (array_key_exists($url, $routes)) {
+                    $routes[$url][$locale] = $routeURI;
+                } else {
+                    $routes[$url] = [$locale => $routeURI];
+                    $routeName = explode('.', $route->getName());
+                    $page = Page::where('slug->en', '=', $routeName[count($routeName) - 1])->first();
+                    if ($page) {
+                        $lastmod[$url] = $page->updated_at->toISOString();
```
jobara: This is interesting, and would work with the current pages we have. I guess the issue is that it's technically possible for there to be more Pages in the future. Perhaps the privacy policy would use this scheme or some other document. At the moment we have removed Creating pages from the filament admin panel, but it is still possible to make them programmatically. Also the route name and slug matching may not always be the case.

Another thing is that we are now running lookups for Pages for every route that matches the patterns.

Let me know what your thoughts are on these. Perhaps this is something we live with for now and file a ticket to further refine in the future, or maybe there are other improvements that can be made now.

jobara (changes requested):

app/Console/Commands/GenerateSitemap.php
jobara: Of I forgot to add that if you keep the above you can probably use `Page::firstWhere` instead of `Page::where()->first()`

jobara:
> Thanks for the suggestion @jobara. I initially went with your suggestion of looping through all `Page` models, but then realized that could be omitted if we could query exact Page model we are interested in. Also, I wasn't too happy about how redundant the code blocks become, so I took a slight different approach. Please let me know if there's any issue you see from my approach.

@chosww I left a [comment](https://github.com/accessibility-exchange/platform/pull/1959#discussion_r1391103509) related to this above.

app/Console/Commands/GenerateSitemap.php
chosww: Thanks @jobara, few thoughts in my mind after reading your comment. We are currently checking pages those are publicly available, those have `.about.` in their route names. Would there be more complex public pages later those require Page model? About the matching route name and slug, since we know that slug is programatically generated using title of specific Page model, we could use route name that follows the slug; if we were to take the approach of using config to tell which route names we want to ignore first and loop through later, we need to update that config value every time we add a page we want to generate sitemap for.

Currently, we do run lookups for Pages for every route, but only for the ones with 'en'. Perhaps I misunderstood your suggested approach of "looping through all `Page` models and getting their urls", but that seemed like more lookup  operations because looping through all Page happens per page in the config; could you explain in details about the approach please?

config/seo.php
```diff
@@ -0,0 +1,24 @@
...
+    |
+    | This file is for storing the configuration values related to search engine optimization.
+    |
+    */
+
+    'sitemap' => [
+        'patterns' => '*.about.*',
+        'to_ignore' => [
+            '*.about.page',
+        ],
```
jobara: This doesn't appear to be used. I suppose we actually wanted to use it to ignore the generic Page route.

app/Console/Commands/GenerateSitemap.php
```diff
@@ -0,0 +1,60 @@
...
+                    $routes[$url][$locale] = $routeURI;
+                } else {
+                    $routes[$url] = [$locale => $routeURI];
+                    if ($route->named(config('seo.sitemap.pages'))) {
+                        $routeName = explode('.', $route->getName());
+                        /*
+                         * TODO: come up with better query for the page, as slug may not follow the pattern in route name
+                         * or changes to the slug would break this logic: https://github.com/accessibility-exchange/platform/issues/1973
+                         */
+                        $page = Page::firstWhere('slug->en', '=', $routeName[count($routeName) - 1]);
```
jobara: I wonder if you can use [`end()`](https://www.php.net/manual/en/function.end.php) here instead.

```suggestion
                        $page = Page::firstWhere('slug->en', '=', end($routeName));
```

jobara (changes requested):

resources/views/sitemap.blade.php
jobara: @chosww I believe you also need to include the en locale in the links. So the conditional. See: https://developers.google.com/search/blog/2012/05/multilingual-and-multinational-site

jobara (approved):

PR #1981, opened 2023-11-16
ci(deploy): add command & queue to put site into maintenance and brin…

### LINKED ISSUE #1867: During rolling deploy it is possible for the old application pod to interact with the updated database
**Prerequisites**

**Describe the bug**

In our current rolling deploy system, as new pods are being deployed an old pod sticks around until the new ones are ready for use. However, there is a single shared database that the pods connect to. The issue here is that a user may be interacting with the old pod, but the database could have been migrated to a new structure. This could lead to data corruption and/or 500 errors reported to the user as the application may have a mismatch of expectations of the data compared to the current database.

**Expected behavior**

We should minimize or eliminate the possibility of the old application and new database from interacting with each other.

marvinroman:
Creates a method to put the site into maintenance mode during delpoyments and creates a delayed queue that will bring the site back up within a certain amount of time (5 minutes).

- Creates new command `maintenance:queue` that initiates putting the site into maintenance mode and initiating the delayed queue.
- Creates new queue **BringSiteBackUp** that brings the site out of maintenance mode.
- Added calling the command `maintenance:queue` at the start of the entrypoint.
- Added running the laravel queue worker for 15 minutes after the pod starts.
- Changed the `QUEUE_CONNECTION` from **sync** to **redis** this allows for delayed queues to occur.

### DIFF
--- a/app/Console/Commands/MaintenanceModeQueueCommand.php
+++ b/app/Console/Commands/MaintenanceModeQueueCommand.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace App\Console\Commands;
+
+use App\Jobs\BringSiteBackUp;
+use Illuminate\Console\Command;
+
+class MaintenanceModeQueueCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'maintenance:queue';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Put the site into maintenance mode and bring it back up after 5 minutes';
+
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        // Put the site into maintenance mode
+        $this->call('down');
+
+        // Schedule bringing the site back up after 5 minutes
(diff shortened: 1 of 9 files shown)


app/Console/Commands/MaintenanceModeQueueCommand.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Execute the console command.
+     */
+    public function handle()
+    {
+        // Put the site into maintenance mode
+        $this->call('down');
+
+        // Schedule bringing the site back up after 5 minutes
+        BringSiteBackUp::dispatch()->delay(now()->addMinutes(5));
```
jobara, line 33: I know you mentioned in your [comment on the ticket](https://github.com/accessibility-exchange/platform/issues/1867#issuecomment-1814977282) that lifecycle hooks aren't possible at the moment in your infrastructure. I'm just wondering if there really isn't a way to know when the old pod is removed and to trigger the `artisan up` call then. Unfortunately adding a delay doesn't adequately address the issue.

marvinroman: We weren't able to come up with an alternative.

PR #1987, opened 2023-11-21

theaccessibilityexchange:


jobara:
After updating from Dev there are no local changes. This is likely do to the translation changes related to #1988 having already included translation updates. These were merged up to Crowdin after this PR was originally created by Crowdin.

PR #2020, opened 2023-11-29
ci: add Release Please (#2019)

### LINKED ISSUE #2019: Configure release-please and (optionally) automatic PRs from dev to staging/production

Currently the release process is largely manual (tagging release, merging changes into staging and production branches).

Add GitHub Actions to:

- Prepare release PRs based on commits to dev (when merged they would create a GitHub release)
- Open PRs from tag to staging whenever a GitHub release is tagged
- Open PRs from tag to production whenever a non-prerelease GitHub release is tagged

N/A.

### Additional context

N/A.

greatislander:
Adds:

- [x] GitHub Action to create a release using Release Please
- [x] GitHub Action to open PRs to `staging` whenever a release is published
- [x] GitHub Action to open PRs to `production` whenever a stable release is published


.github/workflows/sync-to-staging.yml
```diff
@@ -0,0 +1,18 @@
+name: Sync release to staging
+
+on:
+  release:
+    types: [prereleased, released]
```
jobara: There's a note in the [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release):

> Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

I think that means the types here should be switched to `[published]`.

jobara (changes requested):
What do you think of also adding the option to manually trigger the branch syncing actions? This would allow us to trigger the PR creation if the GitHub Actions failed to automatically run, the original PRs were closed in error. 

If you add the [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) event it will add the ability to trigger from CLI or GitHub UI. You can setup inputs for it to take in as well. For example in this case you'd want the tag name.

jobara (approved):

PR #2063, opened 2023-12-11
fix: fonts fail to load on www subdomain (resolves #2006)

### LINKED ISSUE #2006: Fonts fail to load on www subdomain of production site
**Prerequisites**

**Describe the bug**

The fonts do not load when the site is accessed from the www subdomain. It appears they are trying to be accessed from the root domain and through a CORS error.

**To Reproduce**
1. Go to www.accessibilityexchange.ca
2. Notice the fonts don't appear as expected (compare to https://accessibilityexchange.ca)
3. Open the dev tools 
4. Notice the fonts are trying to be accessed from accessibilityexchange.ca and returning a CORS error

**Expected behavior**

**Screenshots**

Screenshot shows fonts not loading due to CORS error

jobara:
Resolves #2006 

The fonts.css file is generated when `php artisan google-fonts:fetch` is called. It is configured to use the `public` disc which had been setup to include the `APP_URL` to generate an absolute URL to the resource. However, because this is configured to be the URL without the sub domain, it produces the wrong URL when the www subdomain is used. To resolve the issue, a URL relative to the root is used instead.

This can be tested locally by adding the www subdomain in front. For example if configured to run on `platform.test` you can use `www.platform.test`. This works with Herd at least, so may depend on your local setup.

### Prerequisites


jobara:
@greatislander could you please review this? I made a change to the public disc, so would like to make sure nothing else will be negatively affected.

greatislander (approved):
Looks good! ✅

PR #2130, opened 2024-02-02

theaccessibilityexchange:


jobara:
This PR was created when there was an issue with the English source translation. Fixing that removed the need for this PR which is why there are no longer any changes.

PR #2152, opened 2024-02-29
fix: split the healthcheck and external Node Ping paths (resolve #2042)

### LINKED ISSUE #2042: Node ping errors nightly on dev during database reset
### Prerequisites

- [X] I've [searched](https://github.com/accessibility-exchange/platform/issues) for existing issues

### Describe the bug

When the dev site's database is reset nightly, around 3am ET, the site is put into maintenance mode. During this period the node pings fail and send out an error report. This is not a true failure as the dev site is expected to be in maintenance mode at this time.

### Expected behavior

There shouldn't be any error alerts if the site is in maintenance mode during this expected time. However, if it isn't removed from maintenance mode in a reasonable amount of time, it should trigger an alert.

### Additional context

See [discussion](https://github.com/accessibility-exchange/platform/issues/1867#issuecomment-1814966474) about the health check on #1867.

The options would seem to be something along the lines of:

- Modify the health check
- Do not put the site into maintenance mode. This could mean that database is reset while someone is in the middle of interacting with it. They'll get errors, at least with their session but possibly also user account, and any data they are trying to save.

marvinroman:
Split the healthcheck and external Node Ping paths that are checked.

Added the custom status pages to be exceptions from maintenance mode so they can be hit by kubernetes even when the site is in maintenance mode.


.kube/app/Dockerfile
```diff
@@ -65,7 +65,7 @@ COPY . $APP_DIR
 
 RUN composer install
 
-RUN . "$NVM_DIR/nvm.sh" && nvm use $NODE_VERSION && npm ci
+RUN . "$NVM_DIR/nvm.sh" && nvm use $NODE_VERSION && npm ci && npm run build
```
jobara: @marvinroman this is interesting because there is a [post install script in the package.json file](https://github.com/accessibility-exchange/platform/blob/4290ace63f908a4b6b20e368454686a82a34ac52/package.json#L14); which should be running `npm run build`. When I run `npm ci` locally, the build is also run. Can you see if the build is run as part of the post install in docker? Maybe something isn't working or maybe something else is at play.

marvinroman: @jobara I have removed. 

I can confirm that the build directory was mapped into the running container which probably caused the anomaly in my local environment.

jobara (approved):

PR #2209, opened 2024-04-19

theaccessibilityexchange:


resources/lang/fr.json
```diff
@@ -348,17 +348,17 @@
     "Communication disabilities": "Troubles de la communication",
     "Communications": "Communications",
     "Communities served": "Communautés desservies",
-    "Communities we :represent_or_serve_and_support": "Communautés que nous :represent_or_serve_and_support",
+    "Communities we :represent_or_serve_and_support": "Communautés que nous représentons",
```
marvinroman: @colinbdclark Should this be replaced this seems like a variable value?

marvinroman:
Merge is approved pending whether `:represent_or_serve_and_support` should be translated.

resources/lang/fr.json
jobara: Depending on the context, `:represent_or_serve_and_support` is replaced by the translations for `__('represent')` or `__('serve and support')`. It's likely that the placeholder should remain.

marvinroman: Thanks @jobara I will go ahead and revert it.
I also excluding language files from Prettier since it will remove `\/` from strings when it formats the file.

marvinroman (approved):

PR #2339, opened 2024-10-29
fix: syntax error when running prettier on blade files (resolves #2338)

### LINKED ISSUE #2338: Syntax error when trying to run prettier on some blade files
### Prerequisites

- [X] I've [searched](https://github.com/accessibility-exchange/platform/issues) for existing issues

### Describe the bug

While working on #2276 there needed to be a change to the resources/views/engagements/leave.blade.php file. When attempting to commit changes an error occurred from prettier about a syntax issue.

```log
✖ prettier --write:
[error] resources/views/engagements/leave.blade.php: SyntaxError: Parse Error : syntax error, unexpected '___attr...' (T_STRING), expecting ';' on line 1
[error] > 1 | <?php echo $engagement->connector->preferred_contact_method ___attrs_5___ &&
[error]     |                                                            ^
[error]   2 |                                 $engagement->connector->contact_phone ?>
[error] > 1 | <x-app-layout page-width="medium">
[error]     |                                   ^
[error]   2 |     <x-slot name="title">{{ __('Leave engagement') }}</x-slot>
[error]   3 |     <x-slot name="header">
[error]   4 |         <h1>
```
This is triggered by the [`lint-staged` task for blade files](https://github.com/accessibility-exchange/platform/blob/8f792d28ade1b6f9d9fdc83b4d2e450b5752d7de/package.json#L28-L30). It's unclear how long this issue has existed because this is only run against files that have been changed and are going to be committed.

### To Reproduce

1. Make any change, including white-space, to the resources/views/engagements/leave.blade.php file
2. Attempt to commit that change
3. Notice the error message

### Expected behavior

Files should be committed as expected.

jobara:
Resolves #2338 

- Fixes syntax issues in blade files that were preventing prettier from running
- Runs prettier on CSS, JS, and Blade files

_**NOTE:** Currently eslint-config-prettier hasn't been updated to support eslint stylistic rules; which means that there is now a conflict between linting and prettier. This is seen in the datePicker.js file for the indentation of the switch/case. For now these are left to work with eslint's rules. See: https://github.com/prettier/eslint-config-prettier/pull/272_

### Prerequisites

### DIFF
--- a/resources/views/engagements/leave.blade.php
+++ b/resources/views/engagements/leave.blade.php
@@ -19,26 +19,30 @@
                                 'support_person_name' => $engagement->connector->user->support_person_name,
                             ]) }}</strong><br />
                         @if ($engagement->connector->contact_email)
-                            <x-contact-point type='email' :value="$engagement->connector->contact_email" :preferred="$engagement->connector->preferred_contact_method === 'email' &&
-                                $engagement->connector->contact_phone" />
+                            <x-contact-point type='email' :value="$engagement->connector->contact_email"
+                                preferred="{{ $engagement->connector->preferred_contact_method === 'email' && $engagement->connector->contact_phone }}" />
                         @endif
                         @if ($engagement->connector->contact_phone)
-                            <x-contact-point type='phone' :value="$engagement->connector->contact_phone" :preferred="$engagement->connector->preferred_contact_method === 'phone' &&
-                                $engagement->connector->contact_email" :vrs="$engagement->connector->contact_vrs" />
+                            <x-contact-point type='phone' :value="$engagement->connector->contact_phone"
+                                preferred="{{ $engagement->connector->preferred_contact_method === 'phone' && $engagement->connector->contact_email }}"
+                                :vrs="$engagement->connector->contact_vrs" />
                         @endif
                     </p>
                 @elseif($engagement->organizationalConnector)
                     <p>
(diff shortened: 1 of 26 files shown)


resources/views/engagements/leave.blade.php
```diff
@@ -19,26 +19,30 @@
                                 'support_person_name' => $engagement->connector->user->support_person_name,
                             ]) }}</strong><br />
                         @if ($engagement->connector->contact_email)
-                            <x-contact-point type='email' :value="$engagement->connector->contact_email" :preferred="$engagement->connector->preferred_contact_method === 'email' &&
-                                $engagement->connector->contact_phone" />
+                            <x-contact-point type='email' :value="$engagement->connector->contact_email"
+                                preferred="{{ $engagement->connector->preferred_contact_method === 'email' && $engagement->connector->contact_phone }}" />
```
greatislander, line 23: Are we sure this works? I think I remember there being a distinction where Blade components couldn't use this syntax but can't find documentation on it now.

greatislander (changes requested):
One question in here, but looks fine otherwise.

resources/views/engagements/leave.blade.php
jobara: I have tried to test all these types of changes and I believe they are working. For example I was able to get the appended `(preferred)` to flip between email and phone by changing the related setting. It's worth testing out at least one of these changes again though. I think maybe org or fro contact might be easier.

greatislander (approved):

PR #2633, opened 2025-04-29
ci: add documentation for running artisan commands in nix environment (resolves #2506)

### LINKED ISSUE #2506: Add documentation for running artisan commands in nix environment
There is an option for using nix-shell to create a local environment for testing. We should document how to run artisan commands for this environment so that things like database migrations can be run there.

marvinroman:
- Adds user documentation for nix shell and aliases that are created when entering the nix shell.
- Improves speed of deployment by changing ownership in copy and rsync stream instead of changing afterwards. Drastically improved local build and deployment speed.
- Added more helpful aliases & bash functions to help with local development.

- [x] Flush local images and volumes and build containers from scratch `dcdn && imgrmp && volrmp && dcupd`.
  - [x] Verify ownership of all the files in the `/app` and `/opt/data` containers `find ./ ! -user www-data` & `find /opt/data/ ! -user www-data`.
  - [x] Verify site comes up and can be browsed without errors.
- [x] Run all aliases locally and verify they work as expected.
- [x] Comment out `USER_ID: $WWWUSER` in `docker-compose.yml` rebuild `dcbp` and bring back up `dcupd`. 
  - [x] Verify that www-data user runs as default id 33 and files have the same ownership.

### DIFF
--- a/.kube/app/Dockerfile
+++ b/.kube/app/Dockerfile
@@ -9,6 +9,8 @@ ENV APP_DIR /app
 ENV KUBE_DIR .kube/app
 
 ARG CIPHERSWEET_KEY
+ARG USER_ID
+RUN [ -z "$USER_ID" ] || usermod -u $USER_ID www-data # If argument passed then swap out numerical id for www-data user to allow for mirroring volumes from local to host
 
 RUN apt-get update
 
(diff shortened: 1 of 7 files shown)


marvinroman:
@jobara Jure reviewed the deployment impacting changes and approved them.

jobara (approved):

PR #2645, opened 2025-05-05
chore: add more local dev aliases

marvinroman:
Added more helpful shell aliases and documentation for Nix Shell non-impacting changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,7 @@ Each time you want to have your terminal environment setup you will want to run 
 | `dcbp` | Builds the `platform.test` service |
 | `dcupd` | Starts containers in detached mode |
 | `dcdn` | Stops and removes containers |
+| `dex [options] <service> <cmd>` | Executes an non-interactive command in a running container |
 | `dexit [options] <service> <cmd>` | Executes an interactive command in a running container |
 | `dexp` | Opens a Bash shell in `platform.test` as user `www-data` |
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -258,7 +260,11 @@ Each time you want to have your terminal environment setup you will want to run
 
 | Command | Description |
 |--------|-------------|
+| `analyze` | Runs `composer analyze` and `vendor/bin/phpstan analyze` inside `platform.test` container |
 | `artisan <command>` | Runs any Laravel Artisan command inside `platform.test` container |
+| `comp <command>` | Runs any Composer command inside `platform.test` container |
+| `comp <command>` | Runs any Composer command inside `platform.test` container |
```
jobara: It looks like this is duplicated. Did you mean to add a different command here?

If you're looking for others to add, other composer commands I typically run are:
- `composer format`
- `composer localize`

marvinroman: I've added those aliases and removed the duplicate line in the readme. 

Those commands since they operate on the code it's best to run them locally instead of the container like most. 

Dependending on the volume mirroring from your dockers I wouldn't want to guarantee they would properly update all the local files and have proper ownership and permissions.

Composer and PHP8.4 are installed when you run `nix-shell` so it will be in your path locally if the aliases are.

If you see any others that should similarly be run locally instead of containers let me know?

jobara (approved):

PR #2654, opened 2025-05-09
ci: clean up workflows

greatislander:


.github/workflows/lighthouse.yml
```diff
@@ -7,12 +7,16 @@ on:
     branches: [dev, staging, production]
 
 jobs:
-  check-diff:
```
jobara, line 10: Was there a need to move this back to a stop instead of a separate job?

greatislander: Yes, it didn't work :)

jobara (approved):

PR #2692, opened 2025-06-10
feat: add Document and Revision models and admin UI

### LINKED ISSUE #2689: Provide an admin interface for uploading/managing files for use on the site

There isn't a way to upload files to be shared from the website.

Provide a filament admin interface that would allow for uploading and managing files that are distributed through the site. Initially this will be for uploading an excel form for WORBE. We'll need to ability to upload and retain different versions of the excel, all of which would be accessible through the site on one or more pages present to the user.

We may want to upload these to an S3(or compatible) storage.

_No response_

### Additional context

The files should be accessible for download without logging in, although there is a request to be able to track download statistics and also capture the e-mail info of who downloads. This doesn't have to be part of this interface, per se, but being able to view this information as part of the interface would be helpful.

greatislander:
Resolves #2689.

### Prerequisites


app/Filament/Resources/RevisionResource.php
```diff
@@ -0,0 +1,93 @@
...
+        ];
+    }
+
+    public static function getFilename(string $language, string $extension, Document $document, ?Revision $revision): string
+    {
+        $name = Str::slug($document->getTranslation('name', $language));
+        $date = $revision ? $revision->created_at->format('Y-m-d') : Carbon::now()->format('Y-m-d');
+
+        return "$name-$date-$language.$extension";
+    }
```
jobara: It seems that this will name revisions based on the slug of the document name and the date (Y-m-d). However, it doesn't account for multiple revisions occurring on the same day. If this happens it will overwrite the previous file. The revision models will still be in tact, but they will be pointing on at the same file on disk. This also poses an issue  when deleting a revision, as it may remove a file that is used by multiple revisions models.

tests/Feature/DocumentTest.php
```diff
@@ -0,0 +1,36 @@
...
+use App\Models\Revision;
+use Illuminate\Support\Carbon;
+
+test('documents can have many revisions', function () {
+    $document = Document::factory()
+        ->has(Revision::factory()->count(3), 'revisions')
+        ->create();
+
+    expect($document->revisions)->toHaveCount(3)
+        ->and($document->revisions->first()->document_id)->toBe($document->id);
```
jobara, line 14: It would be good if this test could also verify the there are 3 created files stored on disk.

jobara: Another test to delete revisions would also be good.

jobara (changes requested):

greatislander:
Updated to enforce unique revision dates.

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
```diff
@@ -0,0 +1,87 @@
...
+            ]);
+    }
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->recordTitleAttribute('create_at')
+            ->columns([
+                Tables\Columns\TextColumn::make('created_at')->label(__('Date'))
+                    ->date('Y-m-d'),
```
jobara: This should probably be the "Revision date"

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
```diff
@@ -0,0 +1,87 @@
...
+                    ->badge()
+                    ->color(fn (string $state): string => $state ? 'success' : 'danger')
+                    ->formatStateUsing(fn (bool $state): string => $state ? __('Yes') : __('No')),
+                Tables\Columns\TextColumn::make('has_french')->label(__('French'))
+                    ->badge()
+                    ->color(fn (string $state): string => $state ? 'success' : 'danger')
+                    ->formatStateUsing(fn (bool $state): string => $state ? __('Yes') : __('No')),
+            ])
+            ->headerActions([
+                Tables\Actions\CreateAction::make(),
```
jobara: Can we add `->createAnother(false)`? At the moment the "Create" button is blocked during the upload process but the "Create and Create Another" one isn't. If you click on this before the upload is completed it will make the revision but not upload the file.

Another option would be to disabled the "Create and create another" button during upload if that's possible.

see: 
- https://filamentphp.com/docs/3.x/actions/prebuilt-actions/create#disabling-create-another
- https://github.com/filamentphp/filament/discussions/6114

jobara (changes requested):

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
jobara: Similarly in the Revisions table it should also be "Revision date".

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
```diff
@@ -0,0 +1,87 @@
...
+                    }),
+            ]);
+    }
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->recordTitleAttribute('create_at')
+            ->columns([
+                Tables\Columns\TextColumn::make('created_at')->label(__('Revision date'))
```
jobara: The revisions table is showing the created date instead of the revision date
```suggestion
                Tables\Columns\TextColumn::make('date')->label(__('Revision date'))
```

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
```diff
@@ -0,0 +1,87 @@
...
+                    ->validationMessages([
+                        'unique' => __('A revision with this date already exists.'),
+                    ])
+                    ->columnSpan(2),
+                Forms\Components\FileUpload::make('file.en')
+                    ->label(__('File (English)'))
+                    ->requiredWithout('file.fr')
+                    ->disk('public')
+                    ->directory('documents')
+                    ->getUploadedFileNameForStorageUsing(function (?Revision $record, TemporaryUploadedFile $file, RelationManager $livewire): string {
```
jobara: `$record` is `null`. When it gets passed to `getFilename` it then defaults to the current date, so all of the files are still saved with the current date.

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
```diff
@@ -0,0 +1,87 @@
...
+    protected static string $relationship = 'revisions';
+
+    public function form(Form $form): Form
+    {
+        return $form
+            ->schema([
+                Forms\Components\DatePicker::make('date')
+                    ->label(__('Revision date'))
+                    ->format('Y-m-d')
+                    ->unique()
```
jobara: So that the record can be updated.
```suggestion
                    ->unique(ignoreRecord: true)
```

app/Filament/Resources/DocumentResource/RelationManagers/RevisionsRelationManager.php
jobara: I haven't yet been able to find a way to access the record from inside this function when called via the relation manager.

jobara (changes requested):

jobara (approved):

PR #2705, opened 2025-06-13
feat: allow public access to resources (resolves #2693)

### LINKED ISSUE #2693: Allow Resources to be accessed without having to log in.

Currently to view resources the user needs to be logged in. However, in the future we want to be able to link to these resources from external sources without requiring login and signup to happen before.

- Remove the authentication requirement for viewing resources.
- May need to update the site navigation to show the resources link all the time
- For things like Training, we'll want the training to be accessible without login up to the point where we need to start tracking things. For example, viewing a module would be blocked because we need to track progress
  - either don't show links to open the module or request for sign up.

_No response_

### Additional context

_No response_

greatislander:
Resolves #2693.

### Prerequisites


chosww:
I think we need to clarify the scope of this work. The issue seems to suggest that we want the trainings to be public until user progress is stored. However, since trainings are composed of modules and courses, are we also planning to make those public as well? The changes to making resources public look good though.

greatislander:
Discussion with @jobara —  courses should be viewable but nothing else (modules, results etc) and the quiz section on the single course page should be hidden for non-authenticated users.

jobara (approved):

PR #2738, opened 2025-06-23
feat: admin interface for tools

### LINKED ISSUE #2734: Create Tool model and administration pages

greatislander:
Adds Tool model with HasMany relationship to Documents, adds Filament UI for adding and editing Tools.

Resolves #2734.

### Prerequisites


database/migrations/2025_06_23_161358_add_tool_id_column_to_documents_table.php
```diff
@@ -0,0 +1,25 @@
...
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('documents', function (Blueprint $table) {
+            $table->foreignId('tool_id')
+                ->nullable()
+                ->constrained()
+                ->onDelete('cascade');
```
jobara: Deleting the tool will delete the associated documents. I'm not sure that this is correct. However, if this is the correct behaviour, there is another issue where deleting the tool deletes the document, but the underlying revisions aren't deleted.

jobara (changes requested):

database/migrations/2025_06_23_161358_add_tool_id_column_to_documents_table.php
greatislander: Oh you're right, it should delete it.

jobara (changes requested):
When trying to delete a tool with an associated document I got the following error:

```
SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`accessibilityexchange`.`documents`, CONSTRAINT `documents_tool_id_foreign` FOREIGN KEY (`tool_id`) REFERENCES `tools` (`id`)) 
```

```sql
delete from `tools` where `id` = 1
```

Reported from 
https://github.com/accessibility-exchange/platform/blob/3a6a90bcdbbe77172bc6ac54b3b1457d21bcf543/app/Http/Middleware/ConfirmLanguage.php#L30

jobara (approved):

PR #2759, opened 2025-07-02
feat: add prompt about what to do next (resolves #2200)

### LINKED ISSUE #2200: Add post getting started prompt for what to do next

After completing the getting started tasks users may not know what to do next. 

Add a post getting started prompt to guide users to what they can do next. For example an individual consultation participant would be prompted to look at available engagements. 

([Figma wireframes](https://www.figma.com/file/5J9uimqjouSxBJW6JR6uZl/Wireframes---ALL?type=design&node-id=378-24425&mode=dev))

Screenshot of post getting started prompt

_No response_

### Additional context

_No response_

greatislander:
Adds prompts for participant or consultant/connector users which are displayed after required tasks have been completed.

### Prerequisites


resources/views/dashboard/getting-started.blade.php
```diff
@@ -1,56 +1,78 @@
...
+                <p class="h3">{{ __('Congratulations, your role has been fully approved.') }}</p>
+                @if ($user->individual->isParticipant() && !($user->individual->isConsultant() || $user->individual->isConnector()))
+                    <livewire:prompt :model="Auth::user()" modelPath="dismissed_browse_engagements_prompt_at"
+                        :heading="__('Find engagements to join')" :interpretationName="__('Find engagements to join', [], 'en')" interpretationNameSpace="getting_started" :description="__(
+                            'As a participant, you can answer surveys, or join focus groups, workshops, and more.',
+                        )"
+                        :actionLabel="__('Browse engagements')" :actionUrl="localized_route('engagements.index')" />
+                @endif
+                @if ($user->individual->isConsultant() || $user->individual->isConnector())
+                    <livewire:prompt :model="Auth::user()" modelPath="dismissed_browse_organizations_prompt_at"
```
jobara: It looks like `dismissed_browse_organizations_prompt_at` may not be wired up to the User model. I don't see it there and am noticing that it doesn't dismiss the prompt properly.

app/Models/User.php
```diff
@@ -78,6 +78,7 @@ class User extends Authenticatable implements CipherSweetEncrypted, FilamentUser
         'oriented_at',
         'suspended_at',
         'dismissed_customize_prompt_at',
+        'dismissed_browse_engagements_prompt_at',
```
jobara: As we're collecting more of these, I'm wondering if they make sense to be in one of those schemaless attributes. Any thoughts on that?

jobara (changes requested):

resources/views/dashboard/getting-started.blade.php
greatislander: I hadn't set it to `fillable` but that's resolved now.

app/Models/User.php
greatislander: I guess we could do that. Which ones would it make sense to move into schemaless?

jobara: For now I was just thinking of the ones that keep track of the states of prompts. 

- `dismissed_customize_prompt_at`
- `dismissed_browse_engagements_prompt_at`
- `dismissed_browse_organizations_prompt_at`

If it makes sense to also do that for other things like the accept states and oriented and approved ones, we could consider that too.

jobara: @greatislander are you still planning to make these model changes in this PR or will that be saved for another one?

greatislander: In this PR.

chosww (changes requested):
The prompt has different padding values for the left and right, which becomes visibly off when the screen gets narrower. We should make those two padding values the same.

image

chosww (approved):

resources/views/dashboard/getting-started.blade.php
```diff
@@ -1,56 +1,77 @@
...
+                            @stack('completed-steps')
+                        </ol>
+                    </x-expander>
+                @endif
+            @else
+                <p class="h3">{{ __('Congratulations, your role has been fully approved.') }}</p>
+                @if ($user->individual->isConsultant() || $user->individual->isConnector())
+                    <livewire:prompt :model="Auth::user()" prompt="dismissed_browse_organizations_prompt_at"
+                        :heading="__('Find organizations to work with')" :interpretationName="__('Find organizations to work with', [], 'en')" interpretationNameSpace="getting_started" :description="__(
+                            'As a community connector, you can find regulated organizations you can work with to help connect them with participants from your community.',
```
jobara: This line always mentions only community connector, even if the role is accessibility consultant. From the email thread the suggestion was to use "As an Accessibility Consultant or Community Connector..."

resources/views/dashboard/getting-started.blade.php
```diff
@@ -1,56 +1,77 @@
...
+                    <x-expander :summary="__('Completed steps')" level="3">
+                        <x-interpretation name="{{ __('Completed steps', [], 'en') }}" namespace="getting_started" />
+                        <ol class="getting-started__task-list stack" role="list">
+                            @stack('completed-steps')
+                        </ol>
+                    </x-expander>
+                @endif
+            @else
+                <p class="h3">{{ __('Congratulations, your role has been fully approved.') }}</p>
+                @if ($user->individual->isConsultant() || $user->individual->isConnector())
```
jobara: I sent this message to Erick:

>Thanks for following up. So view engagements if they have a CP role, but if only AC and/or CC they get Browse Organization options?

and his reply was

> Yes, I think that will make things easier.

That seems opposite of what was done here. But on re-reading the thread, it does seem like what you implemented was what was mentioned before. I'm thinking we leave it this way for now and revisit if they prefer CP prompt to take precedence.

jobara (changes requested):
I got an error after dismissing prompts from a community org.

`Indirect modification of overloaded property App\Models\Organization::$prompts has no effect`

From app / Livewire /Prompt.php : 47

Also the Getting started dialog doesn't go away if there are other prompts outside of the getting started still present, for Orgs and FROs

resources/views/dashboard/getting-started.blade.php
greatislander: Leaving this for now.

resources/views/dashboard.blade.php
```diff
@@ -60,26 +60,40 @@
...
+    @if (
+        !Auth::user()->prompts->dismissed_customize_prompt_at ||
+            (Auth::user()->organization && !Auth::user()->organization->prompts->dismissed_invite_prompt_at) ||
+            (Auth::user()->regulatedOrganization &&
+                !Auth::user()->regulatedOrganization->prompts->dismissed_invite_prompt_at))
+        </div>
+    @endif
 
-    @if ($user->hasTasksToComplete())
+    @if ($user->hasTasksToComplete() || !$user->hasDismissedPrompts())
```
jobara: This will always show the getting started for FROs, Org, and Training users because those fall through to `false` in `$user->hasDismissedPrompts()`.

jobara (changes requested):

resources/views/dashboard/getting-started.blade.php
```diff
@@ -1,56 +1,93 @@
...
+                <ol class="getting-started__list counter stack" role="list">
+                    @if ($user->context === App\Enums\UserContext::Organization->value)
+                        @include('dashboard.partials.getting-started-organization')
+                    @elseif ($user->context === App\Enums\UserContext::RegulatedOrganization->value)
+                        @include('dashboard.partials.getting-started-regulated-organization')
+                    @endif
+                </ol>
+            @else
+                <p class="h3">{{ __('Congratulations, your role has been fully approved.') }}</p>
             @endif
```
jobara: With the latest change the Congratulations message is shown instead of the tasks for the FROs and Orgs. A couple options I can think of is to change the `@else` to an `@elseif` to only display when the `$user` has the individual context. Or add what to do next prompts for Orgs and FROs.

jobara (changes requested):

app/Console/Commands/MigrateSettingsData.php
```diff
@@ -33,6 +34,11 @@ class MigrateSettingsData extends Command implements Isolatable
             'handler' => 'enableEngagementNotificationsMigration',
             'description' => 'Replaces older format of notifications_settings with only ["engagements" => "1"]. Setting the engagement notifications on be default. If the notifications_settings contains a valid engagements setting, then no changes are made.',
         ],
+        'SchemalessPrompts' => [
+            'version' => '1.8.0',
+            'handler' => 'schemalessPromptsMigration',
+            'description' => 'Moves users’ dismissed_customize_prompt_at value to the prompts schemaless attributes column.',
```
jobara: Now that this handles changes for Orgs and Fros the description may need to be updated

jobara:
@greatislander I'm seeing the Getting started panel with just the congratulations message always for Admin and Training users, and for FROs and Orgs after completing all tasks. There isn't a way to remove it either because there is prompt in the getting started like there is for Individual users.

Screenshot of Getting started with just the congratulations message

In these cases the Getting Started shouldn't show at all.

jobara (approved):

PR #2768, opened 2025-07-04
feat: support for adding and editing videos (resolves #2762)

### LINKED ISSUE #2762: Implement an admin interface for site videos

Currently videos used on the site, other than sign language interpretation videos, are hard coded into the blade template files. This makes them more difficult to update.

There should be an admin interface for adding videos that are used on the site. Ideally the pages that incorporate these videos would also be editable but a start would be to make it possible to at least update the URLs for the videos by having database entries for them.

_No response_

### Additional context

_No response_

greatislander:
Heavily based on interpretations functionality.

### Prerequisites


app/View/Components/Video.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public string $videoSrc;
+
+    /**
+     * The identifier name used to reference the videos.
+     */
+    public string $name;
+
+    /**
+     * The explicit namespace to organize the interpretation under
```
greatislander: Update reference here.

app/Models/Video.php
```diff
@@ -0,0 +1,54 @@
...
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Spatie\Translatable\HasTranslations;
+
+/**
+ * App\Models\Video
+ *
+ * @property string $name
+ */
+class Video extends Model
```
jobara, line 15: Video and Interpretation models look basically the same. Is having two only to distinguish the types of videos in filament? If that's the case is there another way to do that distinction, maybe make Interpretation extend Video?

greatislander: I guess we could have done that but now we have fully populated table of interpretations and it doesn't make sense the other way around. We could migrate the data?

jobara: If Interpretation extends Video, would there need to be any data migration? Or were you thinking of another approach?

greatislander: Oh, I see what you mean— just extend the existing model, not consolidate the records. Yes, that makes sense.

jobara (approved):

PR #2779, opened 2025-07-08
fix: individual factory creates two Individual records bound to the same User (resolves #2596)

### LINKED ISSUE #2596: Individual factory creates two Individual records bound to the same user
### Prerequisites

### Describe the bug

When using the Individual Factory it actually creates two Individual records, both of which are bound to the same user. This is because there is a UserObserver that creates a new individual record when a user is created with the `individual` context. The user only supports a `hasOne` relationship to the Individual. This means that one of the Individual records won't be accessible through it. Which causes issues when the Individual Factory is used in tests, which we tend to have to work around by getting the Individual through the user.

### To Reproduce

1. use tinker `php artisan tinker`
2. run `Individual::factory()->create();`
3. look in the database and notice that there are two Individual records created both pointing at the same user id

### Expected behavior

Using the Individual factory should only create the requested number of Individual records, all tied to a unique user record.

### Screenshots

_No response_

### Details

- Device:
- Operating system:
- Browser:
- Link to affected page:

### Additional context

Options to consider for fixing:

- The UserObserver shouldn't create a new Individual record if the user account is already bound to an Individual record
    - The user account may be created before the Individual from the factory. If that's the case, this will be too late to check if a second Individual will be created.
- The individual factory shouldn't create an Individual directly, but rather defer to the one created via a user with the individual context. Then update the individual data with the values passed into the factory, and return the updated individual.

jobara:
Resolves #2596 

- Removes the UserObserver in favour of creating associated Individual model instances during account creation
- Removes duplicate `name` in Individual instance in favour of an accessor that sources the name from the underlying User
- Provides a data migrator to reset the blind_indexes
- Provides a migration to remove the `name` column form the individuals database
- Updates tests
- Refactoring to use enum values instead of hardcoded values
- Updates collections that were returned orderedBy Individual name, to be ordered by the underlying user's name

### Prerequisites


greatislander (changes requested):
Adding this to `src/app/Models/Individual.php` fixes the analysis errors:


```diff
/**
 * App\Models\Individual
 *
+ * @property string $name
 * @property \Spatie\SchemalessAttributes\SchemalessAttributes $extra_attributes
 */
class Individual extends Model implements CipherSweetEncrypted
```

jobara:
> ```diff
> * @property string $name
> ```

Thanks I've added that

app/Console/Commands/MigrateSettingsData.php
```diff
@@ -39,6 +40,11 @@ class MigrateSettingsData extends Command implements Isolatable
             'handler' => 'schemalessPromptsMigration',
             'description' => 'Moves user, regulated organization and organizations’ prompts to the new prompts schemaless attributes column.',
         ],
+        'UpdateBlindIndexes' => [
```
greatislander, line 43: Now that we have this I think we should maybe change this console command to be a bit more generic, e.g. `MigrateData`, as it is not just dealing with settings anymore.

database/migrations/2025_06_12_183542_remove_name_from_individuals_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    /**
+     * Run the migrations.
+     */
```
greatislander: I've been removing these docblocks, along the lines of: https://github.com/spatie/laravel-stubs/blob/main/stubs/migration.stub

greatislander (changes requested):

app/Console/Commands/MigrateSettingsData.php
greatislander: I'm going to file a new issue for this.

greatislander (approved):

Issue #2761, opened 2025-07-02
Uploading an apple numbers spreadsheet saves it as a zip


jobara:
### Prerequisites

- [x] I've [searched](https://github.com/accessibility-exchange/platform/issues) for existing issues

### Describe the bug

If you try to upload an Apple Numbers spreadsheet it is saved as a zip file instead. 

### To Reproduce

1. Login as an admin
2. Go the admin panel and create a new document
3. Add a new revision and upload a Numbers spreadsheet
4. Look at the file system and notice it is saved as a zip file
5. Alternatively add the document to a tool and go the tool page and download the revision.
6. Notice it is zip file downloaded.

### Expected behavior

The same file type that is uploaded should be saved and downloadable. Another alternative is to prevent uploading file types that aren't supported.

### Details

- Device:
- Operating system:
- Browser:
- Link to affected page:

PR #2789, opened 2025-07-11
fix: restrict file upload types ( #2761)

greatislander:
Disallows file uploads other than Excel, PowerPoint, Word and PDF.

### Prerequisites


chosww (changes requested):
Screenshot 2025-07-11 at 12 05 52 PM

The validation error message isn't working properly. Also, should we allow docx format as well?

greatislander:
> Screenshot 2025-07-11 at 12 05 52 PM
> The validation error message isn't working properly. Also, should we allow docx format as well?

That's because the file is too big, which is configured at the server level. Can you try with a very small Word Document (.docx is allowed)?

chosww:
> > > Screenshot 2025-07-11 at 12 05 52 PM
> > > The validation error message isn't working properly. Also, should we allow docx format as well?
> > 
> > That's because the file is too big, which is configured at the server level. Can you try with a very small Word Document (.docx is allowed)?

> Oh I see, small .docx did work. Though, I think we should communicate the size limit somewhere on the UI.

We will once we know what it is :)

PR #2803, opened 2025-07-16
fix: show “My projects” for both orgs and regulated orgs

greatislander:
Both Organizations and Regulated Organizations have consistent navigation to projects they are running in the primary nav.

### Prerequisites


resources/views/components/navigation.blade.php
```diff
@@ -12,10 +12,10 @@
                 </x-nav-link>
             </li>
             @if (Auth::user()->hasVerifiedEmail() && Auth::user()->can('viewOwned', 'App\Models\Engagement'))
-                @if (Auth::user()->context === 'regulated-organization')
+                @if (Auth::user()->context === 'regulated-organization' || Auth::user()->context === 'organization')
```
jobara: This is showing "My projects" for the Community Org, but now they don't see Engagements which they probably still should.

jobara (dismissed):

resources/views/components/navigation.blade.php
```diff
@@ -12,13 +12,15 @@
...
+                @if (Auth::user()->context === App\Enums\UserContext::RegulatedOrganization->value ||
+                        Auth::user()->context === App\Enums\UserContext::Organization->value)
                     <li>
                         <x-nav-link :href="localized_route('projects.my-projects')" :active="request()->localizedRouteIs('projects.my-projects')">
                             {{ __('My projects') }}
                         </x-nav-link>
                     </li>
-                @else
+                @endif
+                @if (!Auth::user()->context === App\Enums\UserContext::RegulatedOrganization->value)
```
chosww: This negation is applied to Auth::user()->context I think, which makes the engagement nav menu to disappear for both individual and organization users. We should wrap the condition before applying the negation, or we could use `Auth::user()->context !== App\Enums\UserContext::RegulatedOrganization->value`

chosww (changes requested):
Screenshot 2025-07-16 at 9 13 36 AM

My Projects tab is showing on the nav bar before Organization or RegulatedOrganization fills in base info about their organization, but when I click it, I see 404 page. I think we shouldn't have a nav item menu to 404 page.

chosww (approved):

PR #2815, opened 2025-07-18
feat: add ContactMethod (fix #1345), use enum values (fix #1343)

### LINKED ISSUE #1343: Values associated with enums should be sourced from the enum as opposed to using raw string values
**Prerequisites**

**Describe the bug**

In the code and blade templates there are times when a check is performed to see if a model attribute is set to some specific value. In the cases where the model attributes are backed by an enum, the comparison should be made against the enum class's value as opposed to a raw string. 

For example: 

https://github.com/accessibility-exchange/platform/blob/461e41d5b89e49de7004bcfc6c824d424335ad0c/resources/views/settings/notifications/individual.blade.php#L33

In the above `$user->preferred_contact_person` relates to an instance of App\Enums\ContactPerson and could have been written like `$user->preferred_contact_person == ContactPerson::Me->value`. 

Additionally, default values and etc should also make use the enums value.

For example:

https://github.com/accessibility-exchange/platform/blob/461e41d5b89e49de7004bcfc6c824d424335ad0c/app/Models/User.php#L49

The above could have been written as `'preferred_contact_person' => ContactPerson::Me->value,`

In doing this, changes to the raw value will not require extensive code changes, and it provides more context to someone reading the code about what the compared value is for.

**Additional context**

In some cases it may be possible to use just the enum case and not having to access its value. e.g. `ContactPerson::Me`.

### LINKED ISSUE #1345: Provide an enum for contact method
**Prerequisites**

**Describe the bug**

Contact method is currently just a string of `'email'` or `'phone`'. Such as:

https://github.com/accessibility-exchange/platform/blob/461e41d5b89e49de7004bcfc6c824d424335ad0c/app/Models/Organization.php#L60

These should be captured in an Enum, to group the possible contact methods together. There is a NotificationMethod Enum already that has values for email, phone, and sms. This could be renamed to ContactMethod and used for both case.

**Additional context**

If the contact method enum includes sms, there may need to be some new inputs for capturing the phone number to use to send the text message to, and updates to the views to display the sms number.

greatislander:
### Prerequisites


app/Http/Controllers/SettingsController.php
```diff
@@ -321,8 +320,6 @@ public function editNotificationPreferences(): View
             'notificationMethods' => Options::forEnum(NotificationMethod::class)->nullable(__('Choose a notification method…'))->toArray(),
             'emailNotificationMethods' => Options::forEnum(NotificationMethod::class)->reject(fn (NotificationMethod $method) => $method === NotificationMethod::Phone || $method === NotificationMethod::Text)->nullable(__('Choose a notification method…'))->toArray(),
             'phoneNotificationMethods' => Options::forEnum(NotificationMethod::class)->reject(fn (NotificationMethod $method) => $method === NotificationMethod::Email)->nullable(__('Choose a notification method…'))->toArray(),
-            'notificationChannels' => Options::forEnum(NotificationChannel::class)->toArray(),
```
greatislander, line 324: These were no longer used in the template so I removed them.

app/Http/Controllers/EngagementController.php
```diff
@@ -107,7 +107,7 @@ public function store(StoreEngagementRequest $request, Project $project): Redire
         flash(__('Your engagement has been created.'), 'success|'.__('Your engagement has been created.', [], 'en'));
 
         $redirect = match ($engagement->who) {
-            'organization' => localized_route('engagements.show-criteria-selection', $engagement),
+            UserContext::Organization->value => localized_route('engagements.show-criteria-selection', $engagement),
```
greatislander: Shouldn't this be `WhoToEngage`?

jobara: @greatislander I believe you're correct. I've made the change.

jobara (approved):

PR #2428, opened 2024-12-10
ci: :construction_worker: changed to building imagick from source

marvinroman:
Updated Dockerfile that builds the main application image to build the imagick library from source instead of using `pecl` since their version is not yet fully ready for php v8.4.

PR #2895, opened 2025-08-25
fix: updates for local nix-shell

greatislander:
This PR adds a new required dependency for the platform image, `libicu-dev`, to address build errors related to `icu-ic`, `icu-io`, and `icu-i18n` not being found (see: https://github.com/docker-library/php/issues/1216). It also restores the PECL `imagick` install as building from source is no longer required (see #2428).

### DIFF
--- a/.kube/app/Dockerfile
+++ b/.kube/app/Dockerfile
@@ -21,6 +21,7 @@ RUN apt-get install -y \
     default-mysql-client \
     git \
     htop \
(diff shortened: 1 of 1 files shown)


marvinroman (approved):
Successfully built with the updates and site appeared to function normally. 
Was able to browse through the site. No errors reported in laravel logs or docker logs.

PR #2904, opened 2025-08-27
feat: adjust max_input_time and max_execution_time

greatislander:
In order to support file uploads, we may need to make some adjustments to `max_input_time` and `max_execution_time` to ensure that larger files can be uploaded. With current configuration on https://dev.accessibilityexchange.ca a 7.5MB file fails to upload in time. I've adjusted the `php.ini` with values that work for me locally which hopefully will resolve the issue.


marvinroman (approved):
Those changes are fine.

PR #3038, opened 2026-03-09
fix: add middleware to maintain locale across livewire requests (resolves #3037)

### LINKED ISSUE #3037: Resource Collection page throws error when searching/filtering content after switching language.
### Prerequisites

### Describe the bug

When admin tries to search or filter resources under a resource collection after changing the platform language to French/LSQ, the page throws a 404 error.

### To Reproduce

1. Go to **Resources and training**  from the admin menu bar
2. Go to **Browse all collections**  
3. choose any Resource collection 
4. Change the platform **language** to French/LSQ 
5. Try to search for a specific resource or try to choose any filter
6. **Error:** 404 page will be shown

### Expected behavior

Searching for resources or applying filters should work correctly regardless of the platform language, without throwing a 404 error.

### Screenshots

After switching the language to **French**

Image

Getting **error** while searching/filtering:

Image

### Details

- Device:
- Operating system:
- Browser:
- Link to affected page:

### Additional context

This is working as intended in some cases.

Image

rvs1212:
When searching or filtering on a page in a non-default locale, Livewire update requests fail with an error for pages that use translatable slugs. This middleware stores the locale in the session on page requests and restores it on Livewire requests.  
Resolves #3037 

### Changes

- Added `ResolveRequestLocale` middleware.

### Prerequisites

### DIFF
--- a/app/Http/Middleware/ResolveRequestLocale.php
+++ b/app/Http/Middleware/ResolveRequestLocale.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+use Livewire\LivewireManager;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Persist and restore the application locale across Livewire requests.
+ *
+ * This middleware stores the locale in the session on page requests and
+ * restores it on Livewire requests.
+ *
+ * @see https://github.com/spatie/laravel-sluggable/discussions/228
+ * @see https://github.com/chinleung/laravel-multilingual-routes/issues/70
+ * @see https://github.com/accessibility-exchange/platform/issues/3037
+ */
+class ResolveRequestLocale
+{
+    public function handle(Request $request, Closure $next): Response
+    {
+        if (app(LivewireManager::class)->isLivewireRequest()) {
+            $locale = session('locale');
+
+            if ($locale && in_array($locale, locales(), false)) {
+                app()->setLocale($locale);
+            }
+        } else {
+            session(['locale' => app()->getLocale()]);
+        }
+
+        return $next($request);
+    }
+}
--- a/bootstrap/app.php
+++ b/bootstrap/app.php
@@ -32,6 +32,7 @@ return Application::configure(basePath: dirname(__DIR__))
         ]);
 
         $middleware->appendToGroup('web', [
+            \App\Http\Middleware\ResolveRequestLocale::class,
             \RalphJSmit\Livewire\Urls\Middleware\LivewireUrlsMiddleware::class,
             \App\Http\Middleware\ConfirmLanguage::class,
         ]);
--- a/tests/Feature/Livewire/LibraryResourcesTest.php
+++ b/tests/Feature/Livewire/LibraryResourcesTest.php
@@ -1,9 +1,100 @@
 <?php
 
 use App\Livewire\LibraryResources;
+use App\Models\Library;
+use App\Models\ResourceCollection;
 use Livewire\Livewire;
+
+use function Pest\Laravel\get;
 
 it('renders successfully', function () {
     Livewire::test(LibraryResources::class)
         ->assertStatus(200);
+});
+
+test('resource collections can be paginated in English', function () {
+    $library = Library::factory()->create([
+        'title' => ['en' => 'Inclusive Employment', 'fr' => 'Emploi Inclusif'],
+    ]);
+
+    $collections = ResourceCollection::factory(11)->create();
+    $library->resourceCollections()->attach($collections);
+
+    /** @var \Illuminate\Pagination\LengthAwarePaginator|null $firstPageResults */
+    $firstPageResults = null;
+
+    Livewire::test(LibraryResources::class, ['library' => $library])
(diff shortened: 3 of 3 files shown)


jobara (commented):
I'm not sure this is the best approach as the middleware will be touched by many if not all of the requests made, whether using livewire or not. There may be unintended side effects.

In looking at this more, I wonder if it's related to https://livewire.laravel.com/docs/3.x/upgrading#localization The docs there say that Livewire 3 doesn't respect the locale prefix anymore. You will need to change the update route. However, some people are reporting some other issues with this. https://github.com/livewire/livewire/discussions/8233

In Livewire 4 it seems you update this in a service provider and also allow for a hash value. https://livewire.laravel.com/docs/4.x/installation

This makes me think you should do the Livewire update and see if the issue still persists and if so, if you can change the update endpoint based on the Livewire 4 docs.

rvs1212:
I checked the Livewire docs and also looked into similar discussions around this. The project uses chinleung/laravel-multilingual-routes, which registers separate routes for each locale instead of using a route group prefix approach. I also tried the setUpdateRoute() option, but even if I follow the approach mentioned in the Livewire docs (https://livewire.laravel.com/docs/3.x/upgrading#localization), it still requires middleware to store and restore the locale. In our case, the locale is dynamic per user, so customizing the update route alone does not fully solve the issue. I believe this would be the same situation even if we upgrade to a newer Livewire version. That is why I thought handling this through middleware might be a safer approach.

app/Http/Middleware/ResolveRequestLocale.php
```diff
@@ -0,0 +1,36 @@
...
+    public function handle(Request $request, Closure $next): Response
+    {
+        if (app(LivewireManager::class)->isLivewireRequest()) {
+            $locale = session('locale');
+
+            if ($locale && in_array($locale, locales(), false)) {
+                app()->setLocale($locale);
+            }
+        } else {
+            session(['locale' => app()->getLocale()]);
```
jobara, line 31: Why is the else condition necessary? Is there another similar issue with a different request?

jobara:
> I checked the Livewire docs and also looked into similar discussions around this. The project uses chinleung/laravel-multilingual-routes, which registers separate routes for each locale instead of using a route group prefix approach. I also tried the setUpdateRoute() option, but even if I follow the approach mentioned in the Livewire docs (https://livewire.laravel.com/docs/3.x/upgrading#localization), it still requires middleware to store and restore the locale. In our case, the locale is dynamic per user, so customizing the update route alone does not fully solve the issue. I believe this would be the same situation even if we upgrade to a newer Livewire version. That is why I thought handling this through middleware might be a safer approach.

Ideally you' probably want to be working with the latest versions to make sure that they haven't addressed the issue. There is also a new release of Laravel out now. If you don't do the update before hand, you'll need to leave a tasks to revisit the issue when you do the upgrades.

app/Http/Middleware/ResolveRequestLocale.php
rvs1212: The else case saves the locale to the session on page requests, so it can be restored on Livewire requests.

rvs1212:
> > I checked the Livewire docs and also looked into similar discussions around this. The project uses chinleung/laravel-multilingual-routes, which registers separate routes for each locale instead of using a route group prefix approach. I also tried the setUpdateRoute() option, but even if I follow the approach mentioned in the Livewire docs (https://livewire.laravel.com/docs/3.x/upgrading#localization), it still requires middleware to store and restore the locale. In our case, the locale is dynamic per user, so customizing the update route alone does not fully solve the issue. I believe this would be the same situation even if we upgrade to a newer Livewire version. That is why I thought handling this through middleware might be a safer approach.
> 
> Ideally you' probably want to be working with the latest versions to make sure that they haven't addressed the issue. There is also a new release of Laravel out now. If you don't do the update before hand, you'll need to leave a tasks to revisit the issue when you do the upgrades.

Right, I'm also thinking to upgrade and check if it can be resolved.

rvs1212:
> > > I checked the Livewire docs and also looked into similar discussions around this. The project uses chinleung/laravel-multilingual-routes, which registers separate routes for each locale instead of using a route group prefix approach. I also tried the setUpdateRoute() option, but even if I follow the approach mentioned in the Livewire docs (https://livewire.laravel.com/docs/3.x/upgrading#localization), it still requires middleware to store and restore the locale. In our case, the locale is dynamic per user, so customizing the update route alone does not fully solve the issue. I believe this would be the same situation even if we upgrade to a newer Livewire version. That is why I thought handling this through middleware might be a safer approach.
> > 
> > 
> > Ideally you' probably want to be working with the latest versions to make sure that they haven't addressed the issue. There is also a new release of Laravel out now. If you don't do the update before hand, you'll need to leave a tasks to revisit the issue when you do the upgrades.
> 
> Right, I'm also thinking to upgrade and check if it can be resolved.

@jobara After discussing with Erick with related issues in platform, I think we can do this fix for now and once we upgrade the package I can revisit this case [#3060](https://github.com/accessibility-exchange/platform/issues/3060) as it is affecting Livewire requests for pages with translatable slugs. Could you please review it?

Issue #1802, opened 2023-06-14
Provide a means for the service org/platform admin to view all user accounts


jobara:
**Is your feature request related to a problem? Please describe.**

At the moment the service org/platform admin's manage account page shows the Individual, Organization and Regulated Organization accounts, but not the base user accounts.  The org/regulated org members and admins don't show up for the service org, but they are visible to the organizations admin.

**Describe the solution you'd like**

Have a view so that the platform admin can view all user accounts on the system. It would be good to also have some service/support features built in to help with some user admin issues.

cherylhjli:
Current designs: https://www.figma.com/file/AXufBnFAVHvGdsxnQMVhNF/The-Accessibility-Exchange-Wireframes?type=design&node-id=3715%3A94295&mode=design&t=GyyWXKwgbOR66LVc-1

Need to update to include all users, including members of organizations.

cherylhjli:
Also include accounts that are part way through creation and site admin.

jobara:
Another possibility is to implement this in a Filament admin page.

PR #3042, opened 2026-03-18
feat: notify platform admins on user creation and deletion (resolves #3040)

### LINKED ISSUE #3040: Platform admin notifications for user registration and account deletion

Currently, platform administrators are not notified when new users register or when existing users delete their accounts. Implementing this feature would provide administrators to monitor users effectively.

Notify platform administrators via email and database notifications whenever a user registers or deletes their account. This should cover all user types, including individuals, training participants, organization or regulated organization creators, and invited members joining organizations. Each notification should include key details such as the user’s name, email, account type, and any associated organization information where applicable.

_No response_

### Additional context

_No response_

rvs1212:
Added notification feature for platform admins when:
- A new individual or training participant registers
- A new organization or regulated organization is created
- A member accepts an invitation to join an organization
- A user deletes their account

Platform admins receive both email and dashboard notifications.

Resolves #3040 

### Prerequisites


app/Observers/UserObserver.php
```diff
@@ -37,5 +41,42 @@ public function created(User $user): void
...
+            UserContext::Organization->value,
+            UserContext::RegulatedOrganization->value,
+        ])) {
+            $name = (new EncryptedField(
+                app(CipherSweetEngine::class),
+                'users',
+                'name'
+            ))->decryptValue($user->name);
+
+            $admins = User::whereAdministrator()->where('id', '!=', $user->id)->get();
```
jobara: I don't think the extra `where` is needed as it appears that this is only run when the `$user` is not an administrator.

app/Observers/UserObserver.php
```diff
@@ -37,5 +41,42 @@ public function created(User $user): void
...
+
+            Notification::send($admins, new NewUserRegistered(
+                userName: $name,
+                userEmail: $email,
+                userContext: $user->context,
```
jobara: This will expose the users name and info outside of TAE's system. Is this information actually necessary?

app/Http/Controllers/InvitationController.php
```diff
@@ -34,9 +42,29 @@ public function create(StoreInvitationRequest $request): RedirectResponse
...
+        $invitationable = $invitation->invitationable;
+        $role = $invitation->role;
+        $inviteeEmail = $invitation->email;
 
         $invitation->accept();
 
+        if ($invitationable instanceof Organization || $invitationable instanceof RegulatedOrganization) {
+            $invitee = $this->retrieveUserByEmail($inviteeEmail);
+
+            if ($invitee) {
```
jobara: Is there a case where this is falsey? Depending on the case there may be something else that should be done.

app/Notifications/NewMemberJoinedOrganization.php
```diff
@@ -0,0 +1,52 @@
...
+        public string $memberName,
+        public Organization|RegulatedOrganization $organization,
+        public string $memberEmail,
+        public string $memberRole,
+        public string $userContext,
+    ) {}
+
+    private function contextLabel(): string
+    {
+        return UserContext::labels()[$this->userContext] ?? $this->userContext;
```
jobara: Realistically there shouldn't be a case where the contextLabel doesn't return a value. I wonder if you want to return something like `__('Unknown')` or leave as `null` in this case and log an error.

app/Notifications/NewMemberJoinedOrganization.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function toArray(): array
+    {
+        return [
+            'title' => __('New member joined organization'),
+            'body' => __(':name has joined :organization as :role.', [
+                'name' => $this->memberName,
+                'organization' => $this->organization->getTranslation('name', locale()),
+                'role' => $this->memberRole,
```
jobara: This should follow a similar pattern as the other notifications. These strings are built up in the notification view component instead. This should just pass along the necessary parts needed by the view. Apart from consistency, I believe this will also mean that localization happens at the time the notification view is rendered, instead of when the notification is generated.

app/Notifications/NewMemberJoinedOrganization.php
```diff
@@ -0,0 +1,52 @@
...
+use Illuminate\Notifications\Messages\MailMessage;
+
+class NewMemberJoinedOrganization extends PlatformNotification
+{
+    public function __construct(
+        public string $memberName,
+        public Organization|RegulatedOrganization $organization,
+        public string $memberEmail,
+        public string $memberRole,
+        public string $userContext,
```
jobara: It's likely you'll need to pass in the actually object instance instead of the string, so that you can pass it along to the notification view. This way it can do the localization at render time.

resources/views/mail/new-member-joined-organization.blade.php
```diff
@@ -0,0 +1,21 @@
...
+</ul>
+
+@component('mail::button', ['url' => localized_route('dashboard')])
+{{ __('Dashboard') }}
+@endcomponent
+
+<p>
+@lang('mail.salutation'),<br>
+{{ config('app.name') }}
+</p>
```
jobara: I think this may already be included in another template. 

https://github.com/accessibility-exchange/platform/blob/bf76f3702e232e92b2431d97516a10c4ef4dcc1e/resources/views/vendor/notifications/email.blade.php#L42-L48

jobara (changes requested):

app/Notifications/NewMemberJoinedOrganization.php
rvs1212: Oke. Modified the approach.

app/Notifications/NewMemberJoinedOrganization.php
rvs1212: sure, modified the approach

app/Http/Controllers/InvitationController.php
rvs1212: There won't be a case where $invitee can be falsey.  The request validation confirms the authenticated user's email matches the invitation email, so the user will exist. Removed that.

app/Observers/UserObserver.php
rvs1212: Modified the email notification content considering privacy

app/View/Components/Notification/NewUserRegistered.php
```diff
@@ -10,8 +11,11 @@ class NewUserRegistered extends Notification
 {
     public function __construct(DatabaseNotification $notification)
     {
-        $this->title = $notification->data['title'];
-        $this->body = $notification->data['body'];
+        $this->title = __('New user registered');
+        $this->body = __('A new user, :name, has registered as :context.', [
+            'name' => $notification->data['user_name'],
+            'context' => UserContext::labels()[$notification->data['user_context']],
+        ]);
```
jobara: I don't think this is actually enough information, for example multiple users could have the same name. Also if you just pass along the email address, the user could change that at any time, so it also may not match. This can be probably be mostly mitigated by passing along the User and displaying the email as it is at render time. In an ideal world, you'd provide a link directly to the user, but there isn't a page for that yet, and is something more along the lines of #1802.

screenshot of multiple users with the same name.

app/Observers/UserObserver.php
```diff
@@ -37,5 +41,40 @@ public function created(User $user): void
...
+                'users',
+                'name'
+            ))->decryptValue($user->name);
+
+            $admins = User::whereAdministrator()->get();
+
+            Notification::send($admins, new NewUserRegistered(
+                userName: $name,
+                userContext: UserContext::from($user->context),
+            ));
```
jobara, line 54: I think this is actually too soon to send, as the user hasn't actually complete the registration process yet. It kind of depends on what you want to do with the information. On one hand it would be useful to know when a user is created, even if they aren't fully registered. However, if the point is to use that information to go into the manage accounts, the user may not show up yet. 

You can replicate this by creating a new individual user, and monitoring the emails in mailpit. You'll get the email notification before the registration process completes. I think right after the password is created. If you look in manage accounts, the user won't be visible. They need to complete the other registration steps before they'll show up in the manage accounts screen.

app/Observers/OrganizationObserver.php
```diff
@@ -0,0 +1,26 @@
...
+
+use App\Enums\UserContext;
+use App\Models\Organization;
+use App\Models\User;
+use App\Notifications\NewOrganizationRegistered;
+use Illuminate\Support\Facades\Notification;
+
+class OrganizationObserver
+{
+    public function created(Organization $organization): void
```
jobara, line 13: This may be too soon, and/or there may be a bug in the manage accounts view. The issue is that the Org does appear in the manage accounts screen but it isn't searchable until after the org has drafted a public page.

The org "Org" is in the manage accounts list.

screenshot showing the Org in the manage accounts list

However, searching for "Org" says there are two results but doesn't render the one we're actually looking for.

screenshot showing the Org not appearing in the search results

After drafting a public page for the Org, it is now searchable.

screenshot showing the Org in the search results

app/Observers/RegulatedOrganizationObserver.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        $creator = User::whereBlind('email', 'email_index', $regulatedOrganization->contact_person_email)->first();
+
+        if ($creator) {
+            $admins = User::whereAdministrator()->get();
+            Notification::send($admins, new NewOrganizationRegistered(
+                creatorName: $creator->name,
+                organization: $regulatedOrganization,
+                userContext: UserContext::from($creator->context),
+            ));
```
jobara: Same notification class is used for both the Organizations and Regulated Organizations. Due to this and the nature of its implementation, it's harder to distinguish the actual type of organization created. The context is included in the app notification but there the title is still just organization. Maybe check with Erick to see if this is sufficient, but you might want to actually title these as Organization and Regulated Organization. If you do make this change, you could either created separate notification classes or you could include logic to distinguish the type based on the `UserContext`.

resources/views/components/notification/new-user-registered.blade.php
```diff
@@ -0,0 +1,4 @@
+<x-notification :notification="$notification">
+    <x-slot name="title">{{ $title }}</x-slot>
+    <x-slot name="body">{{ $body }}</x-slot>
+</x-notification>
```
jobara, line 9: For user types that can be managed, it would be helpful to add an actions slot with a link to the Manage Accounts page. This also applies to organization notifications.

app/View/Components/Notification/NewOrganizationRegistered.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $organizationType = $notification->data['organization_type'];
+        $organization = $organizationType::find($notification->data['organization_id']);
+
+        $this->title = __('New organization registered');
+        $this->body = __('A new organization, :organization, has been registered by :name as :context.', [
+            'organization' => $organization->getTranslation('name', locale()),
+            'name' => $notification->data['creator_name'],
+            'context' => UserContext::labels()[$notification->data['user_context']],
+        ]);
```
jobara: As per the other comment about timing of the notifications for new orgs, if the notice is delayed until after the org has a draft of their public page created, then the organization name here should be a link to their public page.

app/View/Components/Notification/NewMemberJoinedOrganization.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $organizationType = $notification->data['organization_type'];
+        $organization = $organizationType::find($notification->data['organization_id']);
+
+        $this->title = __('New member joined organization');
+        $this->body = __(':name has joined :organization as :role.', [
+            'name' => $notification->data['member_name'],
+            'organization' => $organization->getTranslation('name', locale()),
+            'role' => TeamRole::labels()[$notification->data['team_role']],
+        ]);
```
jobara: The organization name should be a link to the organization's public page.

app/View/Components/Notification/NewUserRegistered.php
jobara: Could also make the user name a link to their public page if one has been created/drafted. 

_**NOTE:** this isn't intended as a comment to address the previous comment above._

app/View/Components/Notification/UserAccountDeleted.php
```diff
@@ -0,0 +1,48 @@
...
+{
+    public function __construct(DatabaseNotification $notification)
+    {
+        $contextLabel = UserContext::labels()[$notification->data['user_context']];
+
+        if (isset($notification->data['organization_id'])) {
+            $organizationType = $notification->data['organization_type'];
+            $organization = $organizationType::find($notification->data['organization_id']);
+
+            $this->body = __(':name (:context) from :organization has deleted their account.', [
```
jobara: I think in this case the context is actually referring to the organization and should be moved after the organization's name instead of the user's name.

jobara: Also may want to make the organization's name a link to the organizations public page.

app/Http/Controllers/InvitationController.php
```diff
@@ -34,9 +43,24 @@ public function create(StoreInvitationRequest $request): RedirectResponse
...
+        if ($invitationable instanceof Organization || $invitationable instanceof RegulatedOrganization) {
+            $invitee = $this->retrieveUserByEmail($inviteeEmail);
+
+            $admins = User::whereAdministrator()->get();
+            /** @var Organization|RegulatedOrganization $invitationable */
+            Notification::send($admins, new NewMemberJoinedOrganization(
+                memberName: $invitee->name,
+                organization: $invitationable,
+                teamRole: TeamRole::from($role),
+            ));
```
jobara: The `NewMemeberJoinedOrganization` is used for both organizations and regulated organizations but doesn't distinguish between them in the notification. As mentioned in another comment for new orgs, you should consider making this distinction to keep things clear on the system. Could either be separate classes or logic in the same class that distinguishes the two.

app/View/Components/Notification/UserAccountDeleted.php
```diff
@@ -0,0 +1,48 @@
...
+                'name' => $notification->data['user_name'],
+                'context' => $contextLabel,
+                'organization' => $organization->getTranslation('name', locale()),
+            ]);
+        } else {
+            $this->body = __(':name (:context) has deleted their account.', [
```
jobara: This is slightly confusing in the context of a user who was part of an organization but was removed from the organization before they deleted their account.

In this case this case the context is still "Community Organization" or "Regulated Organization". Probably best to check with Erick to see how he feels about this.

Steps to reproduce:

1. Create an org or fro
2. Invite a new user to it
3. Create the new user account and accept the invite
4. From the org/fro remove the new user
5. As the invited user, delete their account

app/View/Components/Notification/NewUserRegistered.php
```diff
@@ -0,0 +1,34 @@
...
+use App\View\Components\Notification;
+use Illuminate\Contracts\View\View;
+use Illuminate\Notifications\DatabaseNotification;
+
+class NewUserRegistered extends Notification
+{
+    public function __construct(DatabaseNotification $notification)
+    {
+        $this->title = __('New user registered');
+        $this->body = __('A new user, :name, has registered as :context.', [
```
jobara: I think the grammar is a bit off and will likely need to be toggled based on the context. 

For example:

- 'A new user, :name, has registered as a :context.' (e.g. Training Participant)
- 'A new user, :name, has registered as an :context.' (e.g. Individual)

jobara (changes requested):
I think it would be be helpful

app/Http/Controllers/InvitationController.php
rvs1212: Modified the approach for clarity

app/Observers/OrganizationObserver.php
rvs1212: I believe this might be a bug in the manage accounts page view.  when we create an org/reg org(after adding the org/reg org name the notification email will be send to admin) and in the manage accounts page the org/reg org will be visible. Regarding the search issue, when we search for the org/reg org that you created, some times it shows the total matching count but not actually listing all the matching counts, But upon refreshing the page it will correctly list all the matching org or reg even before the org/reg org page is created. 
manage accounts page search/listing issue need to be checked.

public page can be created at a different time and sending notification after this might be affecting the use case of admin. 

i created a reg org `The Fed Org`  and on searching

Image

upon refreshing the page it lists all the 6 matching results.

Image

app/Observers/RegulatedOrganizationObserver.php
rvs1212: modifed the approach using separate classes for clarity.

app/Observers/UserObserver.php
rvs1212: user will show up in the manage accounts page right after password is added and is searchable. waiting for user to finish all the registration steps might be not the case here, user can finish the steps later on. Once the user info is available in the manage account, admin gets the notification about the user.

app/View/Components/Notification/NewOrganizationRegistered.php
rvs1212: Modified the approach to handle the link if org/reg org creates one. As the primary goal is to notify admin about the new user(org/reg org), ideally user should be finishing the registration process and should be creating public page,  but cannot be guaranteed it will the case always, as they can finish the registration at a different time period.

app/View/Components/Notification/NewUserRegistered.php
rvs1212: Modified the approach

app/View/Components/Notification/UserAccountDeleted.php
rvs1212: I agree, as of now it is handled like that, I checked with Erick about the same and he is okay with that.

app/Observers/OrganizationObserver.php
jobara: @rvs1212 please file an issue for this.

tests/Feature/InvitationTest.php
```diff
@@ -185,3 +198,121 @@
...
+        account: $regOrganization,
+        teamRole: TeamRole::Member,
+    ));
+
+    actingAs($admin)->get(localized_route('dashboard.notifications'))
+        ->assertOk()
+        ->assertSee('New member joined')
+        ->assertSee('Test User')
+        ->assertSee('Test Regulated Org')
+        ->assertSee('Regulated organization')
```
jobara: I believe all the application's strings in the tests should use the localized version. e.g. `__('New member joined')`

tests/Feature/RegistrationTest.php
```diff
@@ -319,3 +323,162 @@
...
+        'locale' => 'en',
+        'name' => 'Test User',
+        'email' => 'test-individual-notify@example.com',
+        'context' => UserContext::Individual->value,
+    ])->post(localized_route('register-store'), [
+        'password' => 'correctHorse-batteryStaple7',
+        'password_confirmation' => 'correctHorse-batteryStaple7',
+        'accepted_terms_of_service' => true,
+        'accepted_privacy_policy' => true,
+    ]);
```
jobara: You may want to look into using a [request-factory](https://github.com/worksome/request-factories) for this. There are some examples in the tests. The current request factories are in tests/RequestFactories

jobara (changes requested):
Could you also update the PR branch from dev?

jobara (approved):

PR #3044, opened 2026-04-07
fix: lower minimum allowed value for participants in engagements (resolves #3043)

### LINKED ISSUE #3043: Lower the minimum allowed value for ideal/minimum participants in engagements

Currently, both the ideal and minimum number of participants for an engagement have a floor of 10. This means if an engagement only needs 1 or 2 participants, there's no way to set that as the system requires at least 10. This doesn't reflect properly on engagements where only a small number of participants are needed.

Image

Need to lower the minimum allowed value for both ideal and minimum participants from 10 to 1. Both fields need to change because the ideal participants value is what the system uses to control capacity as it determines whether new participants can join and when the engagement shows as full.

_No response_

### Additional context

_No response_

rvs1212:
- Lowered the minimum allowed value for ideal and minimum participants in engagements.

Resolves #3043

### Prerequisites


config/engagement.php
```diff
@@ -0,0 +1,17 @@
...
+    |--------------------------------------------------------------------------
+    | Minimum Participants Floor
+    |--------------------------------------------------------------------------
+    |
+    | The minimum allowed value for both ideal and minimum participant counts
+    | when creating or updating an engagement.
+    |
+    */
+
+    'minimum_participants_floor' => 1,
```
jobara: Here are a few comments about the configuration:

-  The minimum participants and ideal participants represent different things, so should have different configuration options. 
- This is a step up from being hardcoded throughout the codebase; however, it would still require a change in the source to modify the value.
  - Adding an Environment variable would at least allow it to be configured in the deployment
  - It would be better to have it configurable in the admin panel, maybe something like the site settings.
  - As I mentioned in the issue, it would be better if this could be set on a case-by-case basis too, instead of globally.
- If this is always going to be 1, it makes me wonder if the workflow still makes sense.
  - And this may impact diversity in the engagement.

resources/lang/fr.json
```diff
@@ -1681,7 +1681,7 @@
     "The language you select for navigating this website sets the website into that language.": "La langue que vous sélectionnez pour naviguer le site Internet permettra d’afficher le site dans cette langue.",
     "The language you select for using this website sets the website into that language.": "La langue dans laquelle vous désirez utiliser le site Internet.",
     "The language you want to use for navigating this website.": "La langue que vous souhaitez utiliser pour naviguer sur ce site Internet.",
-    "The least number of participants you can have to go forward with your engagement. The least you can select is 10 participants.": "Il s’agit du nombre minimum de personnes participantes que vous pouvez sélectionner pour mener à bien votre consultation. Le minimum que vous pouvez sélectionner est de 10 personnes.",
+    "The least number of participants you can have to go forward with your engagement. The minimum you can select is :count participant.": "",
```
jobara: You should use the [pluralization](https://laravel.com/docs/12.x/localization#pluralization) feature of the strings so that you can have "participant" or "participants" depending on the count.

jobara (changes requested):

config/engagement.php
rvs1212: I will make the config values separate and add it in env as well. 
Based on the requirement changes, doing case-by-case or configuring separately in admin site settings might not be necessary.

rvs1212:
@jobara could you please re run the checks?

.env.example
```diff
@@ -65,3 +65,6 @@ DOCUMENTS_AWS_DEFAULT_REGION="ca-central-1"
 DOCUMENTS_AWS_BUCKET=
 DOCUMENTS_AWS_ENDPOINT=
 DOCUMENTS_AWS_URL=
+
+ENGAGEMENT_IDEAL_PARTICIPANTS_FLOOR=1
+ENGAGEMENT_MINIMUM_PARTICIPANTS_FLOOR=1
```
jobara: Missing a new line at the end of the file. Also these probably also need to go into the other example env files.

jobara:
> @jobara could you please re run the checks?

It looks like there's an issue with the CipherSweet key. This gets added to the .env file (copied from the example file in CI). I wonder if there is an issue with the changes to the .env.example file.

.env.example
jobara: I did some tests out locally and the with the checks is related to this missing newline. Basically the ciphersweet key is getting appended to the end of the `ENGAGEMENT_MINIMUM_PARTICIPANTS_FLOOR=1`, instead of being on its own line.

resources/views/engagements/show-criteria-selection.blade.php
```diff
@@ -277,10 +277,21 @@ class="sr-only">{{ __('Disability or Deaf group') }}</span>
...
+                                'This is the ideal number of participants you would like to have for this engagement. The minimum you can select is :count participant.',
+                                ['count' => config('engagement.ideal_participants_floor')]
+                            ) .
+                                '|' .
+                                __(
+                                    'This is the ideal number of participants you would like to have for this engagement. The minimum you can select is :count participants.',
+                                    ['count' => config('engagement.ideal_participants_floor')]
+                                ),
+                            config('engagement.ideal_participants_floor')
+                        ) }}
```
jobara, line 284: I think this can actually be written like: 

```suggestion
                        {{ trans_choice(
                                '{1} This is the ideal number of participants you would like to have for this engagement. The minimum you can select is :count participant.|[2,*] This is the ideal number of participants you would like to have for this engagement. The minimum you can select is :count participants.',
                                config('engagement.ideal_participants_floor'),
                                ['count' => config('engagement.ideal_participants_floor')]
                        ) }}
```

Similar for the other case where `trans_choice` is used.

jobara (changes requested):

resources/views/engagements/show-criteria-selection.blade.php
rvs1212: Applied your suggestion for the cleaner trans_choice pattern. I also added trans_choice to the functions array in config/localizator.php so composer localize can pick up pluralized strings passed directly to trans_choice()

config/engagement.php
jobara: I filed a new issue, #3045, for improving the configuration.

jobara (approved):

PR #3052, opened 2026-04-23
fix: deployment timeout failure related to Migrate Data (resolves #3051)

### LINKED ISSUE #3051: Deployment fails due to timeout related to Migrate Data
### Prerequisites

### Describe the bug

When doing a deploy some steps are run once for all pods, and others are done in each pod. In the entrypoint.sh file `php artisan app:migrate-data` is attempted to be run.

Migrate Data is an extra migration step use to update data that a standard migration cannot be used for. While these shouldn't break existing data even when run multiple times, there are now a large number of users in Production and as a result the process appears to be timing out.

This isn't immediately noticeable as the other pod loads, so the site is accessible. However, all of the deploy steps may not have completed so there may be some issues with the site that aren't readily apparent.

### Expected behavior

All of the deployment steps should complete successfully and all pods running properly.

### Screenshots

_No response_

### Details

- Device:
- Operating system:
- Browser:
- Link to affected page:

### Additional context

_No response_

jobara:
Resolves #3051

- Adds more additional verbose outputs to Migrate Data
- Uses verbose output when running `app:migrate-data` from the entrypoint script
- Adds an environment variable to set the from version with when run from the entrypoint script
  - currently set to 1.8.0
- Makes the `app:migrate-data` call from the entrypoint script async 

### Prerequisites

- [ ] I've run `php artisan test` and ensured that all tests pass.
- [ ] I've run `composer localize` to update localization source files and committed any changes.

### DIFF
--- a/app/Console/Commands/MigrateData.php
+++ b/app/Console/Commands/MigrateData.php
@@ -20,7 +20,7 @@ class MigrateData extends Command implements Isolatable
      */
     protected $signature = 'app:migrate-data
                             {--list : lists out available migrations}
-                            {--from=1.6.0 : when running all migrations, indicate which version the application is being migrated from. Previous migrations will be skipped.}
+                            {--from=1.8.0 : when running all migrations, indicate which version the application is being migrated from. Previous migrations will be skipped.}
                             {--migration= : a specific migration to run}';
 
     /**
@@ -95,7 +95,7 @@ class MigrateData extends Command implements Isolatable
         }
     }
 
-    public function runMigrations($migrations, $from = '1.6.0', $verbose = false)
+    public function runMigrations($migrations, $from = '1.8.0', $verbose = false)
     {
         $from = str_starts_with($from, 'v') || str_starts_with($from, 'V') ? substr($from, 1) : $from;
         $migrationRunCount = 0;
--- a/tests/Feature/Console/MigrateDataTest.php
+++ b/tests/Feature/Console/MigrateDataTest.php
@@ -19,7 +19,7 @@ test('app:migrate-data command runs successfully', function () {
 });
 
 test('app:migrate-data command can list migrations', function () {
-    $this->artisan('app:migrate-data --list')
+    artisan('app:migrate-data --list')
         ->assertSuccessful()
         ->expectsOutputToContain('Version added');
 
@@ -43,7 +43,7 @@ test('app:migrate-data command can run migrations starting at a specific version
 });
 
 test('app:migrate-data command can run with verbose logging', function () {
-    artisan('app:migrate-data -v')
+    artisan('app:migrate-data --from=1.6.0 -v')
         ->assertSuccessful()
         ->expectsOutputToContain('Run migration -');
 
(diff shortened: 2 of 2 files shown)


jobara:
From @JureUrsic 

>I'm not really sure introducing MIGRATE_DATA_fROM (which btw has a typo - lower f) is a good way to do data checkpointing, maybe something more dynamic in the code + db would be more appropriate, db migration frameworks usually have a data_version value stored somewhere in db, if that would be too much of a pain to implement I'm thinking maybe reading the git tag and passing it as a --version flag to the script and then do conditional behaviour checking if version  === "1.8.0" or something in that sense...

jobara:
> From @JureUrsic
> 
> > I'm not really sure introducing MIGRATE_DATA_fROM (which btw has a typo - lower f) is a good way to do data checkpointing, maybe something more dynamic in the code + db would be more appropriate, db migration frameworks usually have a data_version value stored somewhere in db, if that would be too much of a pain to implement I'm thinking maybe reading the git tag and passing it as a --version flag to the script and then do conditional behaviour checking if version  === "1.8.0" or something in that sense...

This is a good point. It doesn't look like we have access to the git repo in the deployment. The trick too is that we need to know what we're moving from. Perhaps your point about storing the version in the database would make sense for that. Although this might be a problem for the next deploy if we don't have the value in there yet.

rvs1212:
> > From @JureUrsic
> > > I'm not really sure introducing MIGRATE_DATA_fROM (which btw has a typo - lower f) is a good way to do data checkpointing, maybe something more dynamic in the code + db would be more appropriate, db migration frameworks usually have a data_version value stored somewhere in db, if that would be too much of a pain to implement I'm thinking maybe reading the git tag and passing it as a --version flag to the script and then do conditional behaviour checking if version  === "1.8.0" or something in that sense...
> 
> This is a good point. It doesn't look like we have access to the git repo in the deployment. The trick too is that we need to know what we're moving from. Perhaps your point about storing the version in the database would make sense for that. Although this might be a problem for the next deploy if we don't have the value in there yet.



I agree the database-backed approach seems like the cleaner long-term solution, but it probably needs a bit more work for the initial setup. I think we would need to seed the existing migrations as completed, assuming they have already been applied in the target environments. 
On a different note, would it be worth using `chunkById()` instead of `User::all()` and `Individual::all()` in `updateBlindIndexes()`? If this migration ever needs to run again against a large dataset?.

jobara:
> I agree the database-backed approach seems like the cleaner long-term solution, but it probably needs a bit more work for the initial setup. I think we would need to seed the existing migrations as completed, assuming they have already been applied in the target environments. On a different note, would it be worth using `chunkById()` instead of `User::all()` and `Individual::all()` in `updateBlindIndexes()`? If this migration ever needs to run again against a large dataset?.

Chunking seems like a good idea and may apply to other areas as well. I've filed #3057 to review the entire codebase for areas where we may benefit from chunking.

jobara:
After discussing with @rvs1212 and Erick, we've decided to make a simple fix for now that will just bypass the old data migrations. A new issue, #3058, has been filed to implement a more robust solution.

rvs1212 (approved):

jobara:
@rvs1212 I've modified these changes to just switch the default for data migrations to 1.8.0 which will skip all of the current ones. Could you please review these changes?

rvs1212 (approved):
I believe you removed the async processing and just modified the version to skip the migrations.

jobara:
> I believe you removed the async processing and just modified the version to skip the migrations.

Yes that's pretty much it. I removed all my previous changes and just modified the default from version.

PR #3069, opened 2026-06-16
feat: upgrade Filament to v5 and Livewire to v4 (resolves #3067)

rvs1212:
Upgrade Filament to v5 and Livewire to v4.

- Need to run `composer install` and `php artisan optimize:clear` to test.

resolves #3067 

### Prerequisites


jobara:
@rvs1212 mostly looks good. Did you also go over changes needed for [upgrading livewire](https://livewire.laravel.com/docs/4.x/upgrading)? There seems to have been lots that changed in livewire related to `wire:model`. You may also want to see if it makes sense to use the CSP-safe mode.

Additionally I was just poking around at different parts and noticed that I couldn't save my changes to the "Communication and consultation preferences" for the "Individual User". I'm not sure if this is related to the upgrade, there are some `wire:model` usages on that page, or if it's from something else. Basically I would get a form validation error even when all the required fields were filled. 

Screenshot of form validation error

rvs1212:
> @rvs1212 mostly looks good. Did you also go over changes needed for [upgrading livewire](https://livewire.laravel.com/docs/4.x/upgrading)? There seems to have been lots that changed in livewire related to `wire:model`. You may also want to see if it makes sense to use the CSP-safe mode.
> 
> Additionally I was just poking around at different parts and noticed that I couldn't save my changes to the "Communication and consultation preferences" for the "Individual User". I'm not sure if this is related to the upgrade, there are some `wire:model` usages on that page, or if it's from something else. Basically I would get a form validation error even when all the required fields were filled.
> 
> Screenshot of form validation error


@jobara  Yes, I went through the Livewire v4 upgrade guide and checked the changes against our project. For the `wire:model `changes specifically, the main one is `.blur`, `.change` now also delaying the client-side update. We only use `.blur` on a couple of inputs where nothing reads the value while typing, so the behaviour is the same as before. The other `wire:model` changes only apply to cases we don't have.

From my understanding, I don't think we are enforcing CSP in the project right now, I am not sure just by enabling CSP-safe mode makes a difference. As the platform grows, enforcing  CSP along with enabling CSP-safe mode in Livewire is probably worth considering from a security standpoint.

Regarding the error you got, I could not recreate the issue in both local or staging, it is working as intended when I checked, but I'll keep an eye on that.

jobara:
@rvs1212 I was able to reproduce the issue both locally and on dev (I used the Individual user). I've written up a new issue for this #3072. 

As this was reproduced on dev, it isn't an issue with this PR. I've approve the PR, but will leave it for you to merge when you are ready.

